Issue #3, opened 2016-02-16
Make sure config.json contains all the needed variables for new OS


agherzan:
There were a couple of attributes added to config.json. Ex. vpnEndpoint. Without this variable the hostOS will lose VPN connection. **Make sure all the attributes are defined in the json file. **

PR #4, opened 2016-02-19
Verify config.json and add all the missing info

agherzan:
Verify config.json and add all the missing info there based on running flag

Based on --staging argument (default to production when not present),
config.json will be check for all the required attributes and fix accordingly.

CC @petrosagg @telphan @floion

PR for #3

[Review on Reviewable](https://reviewable.io/reviews/resin-os/resinhup/4)

### DIFF
--- a/conf/resinhup.conf
+++ b/conf/resinhup.conf
@@ -46,7 +46,7 @@ remote: https://s3.amazonaws.com/resin-production-hup/
 workspace: /tmp/resinhup
 update_file_fingerprints: ./resin-boot ./resin-root
 
-[production]
+[staging]
 apiEndpoint: https://api.resinstaging.io
 listenPort: 48484
 mixpanelToken: cb974f32bab01ecc1171937026774b18
@@ -56,7 +56,7 @@ registryEndpoint: registry.resinstaging.io
 vpnEndpoint: vpn.resinstaging.io
 vpnPort: 443
 
-[staging]
+[production]
 apiEndpoint: https://api.resin.io
 listenPort: 48484
 mixpanelToken: 99eec53325d4f45dd0633abd719e3ff1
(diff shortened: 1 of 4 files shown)


conf/resinhup.conf
```diff
@@ -45,3 +45,23 @@ to_keep_files:
 remote: https://s3.amazonaws.com/resin-production-hup/
 workspace: /tmp/resinhup
 update_file_fingerprints: ./resin-boot ./resin-root
+
+[production]
```
petrosagg: The labels are the other way around

agherzan: @petrosagg fixed. Thanks for this.

modules/util.py
```diff
@@ -350,3 +377,26 @@ def mcopy(dev, src, dst):
         log.debug("Failed to mcopy in " + dev);
         return False
     return True
+
+def jsonAttributeExists(json, attribute):
```
petrosagg: Why are we using `jq` for reading and writing JSON? https://docs.python.org/3/library/json.html

agherzan: :) I will use the library. I thought you wanted to use bash commands as much as possible.

agherzan: Implemented in the new PR version. Thanks.

modules/util.py
```diff
@@ -350,3 +378,46 @@ def mcopy(dev, src, dst):
...
+            return False
+        else:
+            log.warn("jsonSetAttribute: " + attribute + " will be overwritten.")
+
+    configjson[atttribute] = value # Set the required attribute
+
+    # Write new json to a tmp file
+    try:
+        with open(json+'.hup.tmp', 'w') as fd:
+            configjson = json.dump(configjson, fd)
```
petrosagg: missing `fsync` call

agherzan: Added. Thanks.

petrosagg:
SGTM

Issue #2, opened 2016-02-16
Implement atomic file operations


agherzan:
shutil.copy is not an atomic operation which might bring the board into an unstable state in the case of a sudden power drop.

PR #5, opened 2016-02-24
Implement atomic copy operations

agherzan:

@floion @petrosagg

PR for #2 

[Review on Reviewable](https://reviewable.io/reviews/resin-os/resinhup/5)


agherzan:
After a talk with @petrosagg : "create tmpfile, copy data, fsync, rename, sync" as a python function and don't use rsync (as rsync dosn't fsync).

modules/util.py
```diff
@@ -350,3 +352,138 @@ def mcopy(dev, src, dst):
...
+        elif os.path.isdir(dst):
+            log.error("Destination target " + dst + " is a directory.")
+            return False
+        else:
+            log.error("Destination target " + dst + " is unknown.")
+            return False
+
+    # Copy file to src.tmp
+    srcfd = open(src, 'rb')
+    srctmpfd = open(src + ".tmp", "wb")
```
petrosagg: These two should be in a `with` block to have python call `.close()` automatically at the end even if there is an exception

petrosagg: The temporary file should be created at the destination, not as a sibling to the source. Otherwise the `rename()` won't work because you'll be crossing a filesystem boundary

modules/util.py
```diff
@@ -350,3 +352,138 @@ def mcopy(dev, src, dst):
...
+    while 1:
+        buffer = srcfd.read(buffer_size)
+        if not buffer:
+            break
+        else:
+            srctmpfd.write(buffer)
```
petrosagg: What happens if less than `buffer_size` bytes were written? Is there some other function of the python library that we can use that takes all these into account?

petrosagg: Is this helpful? https://docs.python.org/3/library/shutil.html#shutil.copyfileobj

modules/util.py
agherzan: Addressed both comments in the new PR version. Thank you.

modules/util.py
agherzan: Addressed comment in the new PR version. Thank you.

petrosagg:
SGTM

PR #8, opened 2016-03-17
Various Fixes

agherzan:
@petrosagg

---
This change is [Reviewable](https://reviewable.io/reviews/resin-os/resinhup/8)

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -5,4 +5,7 @@ set -e
 echo "Stopping all containers"
 rce stop $(rce ps -a -q)
 
+echo "Stop cron jobs"
+/etc/init.d/crond stop
+
 echo "Done"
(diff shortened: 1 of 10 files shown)


prepare.sh
```diff
@@ -5,4 +5,7 @@ set -e
 echo "Stopping all containers"
 rce stop $(rce ps -a -q)
 
+echo "Stop cron jobs"
+/etc/init.d/crond stop
```
petrosagg: This doesn't exist in newer OS versions and will cause this script to fail due to `set -e`

agherzan: Correct. I will remove `set -e.`

agherzan: Addressed in the repartitioner PR.

PR #9, opened 2016-03-19
Repartition mechanism for resin-boot

agherzan:
@petrosagg

---
This change is [Reviewable](https://reviewable.io/reviews/resin-os/resinhup/9)


modules/updater.py
```diff
@@ -336,14 +336,9 @@ def upgradeSystem(self):
...
             return False
-        # Configure bootloader to use the updated rootfs
-        if runningDevice(self.conf) == 'raspberry-pi' or runningDevice(self.conf) == 'raspberry-pi2':
-            b = BCMRasberryPiBootloader(self.conf)
-            if not b.configure(getRootPartition(self.conf), self.toUpdateRootDevice()[0]):
-                log.error("Could not configure bootloader.")
-                return False
-        else:
-            log.warn("No bootloader configuration support for this board.")
+        if not configureBootloader(getRootPartition(self.conf), self.toUpdateRootDevice()[0]):
```
petrosagg: `configureBootloader` takes 3 parameters, only 2 given here

modules/repartitioner.py
```diff
@@ -0,0 +1,223 @@
...
+                # State A
+                #
+                log.debug("Running transition from State A...")
+
+                # Edit resin-updt partition
+                if not self.editPartition(targetPartition=resinUpdtPart, deltaStart=(deltasize // 2), deltaEnd=0, fstype='ext3', fslabel='resin-updt', unit=unit):
+                    log.error("increaseResinBootTo: Could not edit resin-updt partition.")
+                    return False
+
+                # Copy resin-root to resin-boot
```
petrosagg: You mean resin-root to resin-updt

modules/repartitioner.py
```diff
@@ -0,0 +1,223 @@
...
+                log.error("increaseResinBootTo: Unknown filesystem state where booted from resin-boot but having different size then resin-updt.")
+                return False
+
+        elif self.currentResinRootPartPath == resinUpdtPart.path:
+            # Booted from resin-updt
+            if resinRootPart.getLength(unit) == resinUpdtPart.getLength(unit):
+                #
+                # State B
+                #
+                log.debug("Running transition from State B...")
```
petrosagg: Don't you have to copy the contents of `resin-updt` to `resin-root` here before rebooting to it?

modules/util.py
```diff
@@ -62,6 +62,22 @@ def getBootPartition(conffile):
         return bootdevice
     return None
 
+def getPartitionRelativeToBoot(conffile, label, relativeIndex):
+    ''' Returns the partition device path when index is relative to boot partition '''
+    # First search by label
+    partdevice = getDevice(label)
+    if not partdevice:
+        match = re.match(r"(.*?)(\d+$)", getBootPartition(conffile))
```
petrosagg: Why not use pyparted instead of matching over command output?

modules/updater.py
agherzan: Fixed.

modules/repartitioner.py
agherzan: Correct. Fixed.

modules/repartitioner.py
agherzan: Hmm... missed this as I didn't test it. Very good point.

modules/util.py
agherzan: How exactly done with pyparted? I explained in the commit log why I didn't use nextPartitions(). I could use the partitions field of the object but given the fixed setup of our partition table I found it easier and simpler to understand.

petrosagg: How about this? http://pyparted.sourcearchive.com/documentation/2.0.12-1/classparted_1_1disk_1_1Disk_d43678c0a74a3905997cecc68a8a34ea.html#d43678c0a74a3905997cecc68a8a34ea

agherzan: That's exactly the same as partitions attribute. Do you find anything wrong with this approach? I agree it can be done with partitions list too.

modules/repartitioner.py
```diff
@@ -64,6 +90,24 @@ def editPartition(self, targetPartition, deltaStart, deltaEnd, fstype, fslabel,
...
                 return False
 
+        # Restore data in targetPartition
+        if safeDataThroughTmp:
+            # Make sure targetPartition is mounted in mountpoint
+            if isMounted(targetPartition.path):
+                log.error("editPartition: Something is wrong. %s should not be mounted." % targetPartition.path)
+                return False
+            else:
+                if not mount(targetPartition.path, mountpoint):
```
petrosagg: Is there an `else` case missing here?

agherzan: No. If safeDataThroughTmp is not provide the backup block is just ignored.

petrosagg: ok. I got confused because of the extra indentation. if could have been `elif not...`

agherzan: I got it. I just let the code flow if not. Cause I have another block before which does the backup.

resinhup.py
```diff
@@ -111,17 +112,26 @@ def main():
         if not setConfigurationItem(args.conf, "config.json", "type", "production"):
             return False
 
+    # Handle old boot partitions
+    r = Repartitioner(args.conf)
+    if not r.increaseResinBootTo(40):
```
petrosagg: Are our current OSes exactly 40MB? i.e, is this a no-op?

agherzan: Yes - out current boot partition is 40 MiB. I didn't tested on a updated device but should not run on them. Will test when have a chance.

petrosagg:
SGTM

PR #16, opened 2016-04-27
Support for additional devices and implement ignored files in boot partition

agherzan:
@petrosagg @floion @telphan

### DIFF
--- a/Dockerfile.intel-nuc
+++ b/Dockerfile.intel-nuc
@@ -1,0 +1,9 @@
+FROM resin/nuc-debian:jessie
+
+# Install the dependencies
+RUN apt-get update && apt-get install vim btrfs-tools mtools dosfstools python3 python3-requests python3-sh python3-parted python3-binaryornot jq -y && rm -rf /var/lib/apt/lists/*
+
+# Add the current directoy in the container
+ADD . /app
+
+CMD python3 /app/resinhup.py --config /app/conf/resinhup.conf --debug
--- a/Dockerfile.intel-nuc
+++ b/Dockerfile.intel-nuc
@@ -4,6 +4,6 @@ FROM resin/nuc-debian:jessie
 RUN apt-get update && apt-get install vim btrfs-tools mtools dosfstools python3 python3-requests python3-sh python3-parted python3-binaryornot jq -y && rm -rf /var/lib/apt/lists/*
 
 # Add the current directoy in the container
-ADD . /app
+COPY . /app
 
 CMD python3 /app/resinhup.py --config /app/conf/resinhup.conf --debug
--- a/Dockerfile.raspberrypi3
+++ b/Dockerfile.raspberrypi3
@@ -1,4 +1,4 @@
-FROM resin/rpi-raspbian:jessie
+FROM resin/raspberrypi3-debian:jessie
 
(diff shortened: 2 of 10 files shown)


Dockerfile.raspberrypi3
```diff
@@ -0,0 +1,9 @@
+FROM resin/rpi-raspbian:jessie
```
Page-: This would be better as `resin/raspberrypi3-debian:jessie`

modules/bootconf.py
```diff
@@ -14,7 +14,7 @@
 
 def configureBootloader(old, new, conffile):
     ''' Configure bootloader to use the updated rootfs '''
-    if runningDevice(conffile) == 'raspberry-pi' or runningDevice(conffile) == 'raspberry-pi2':
+    if runningDevice(conffile) == 'raspberry-pi' or runningDevice(conffile) == 'raspberry-pi2' or runningDevice(conffile) == 'raspberrypi3':
```
Page-: Can this be `if runningDevice(conffile) in [ 'raspberry-pi', 'raspberry-pi2', 'raspberrypi3' ]`?

Dockerfile.raspberrypi3
agherzan: Correct.

modules/bootconf.py
agherzan: Indeed. Why it happened to get this way is because I added the changes incrementally so or came more handy. :) I'll change.

modules/bootconf.py
```diff
@@ -14,11 +14,16 @@
 
 def configureBootloader(old, new, conffile):
     ''' Configure bootloader to use the updated rootfs '''
-    if runningDevice(conffile) == 'raspberry-pi' or runningDevice(conffile) == 'raspberry-pi2':
+    if runningDevice(conffile) in ['raspberry-pi', 'raspberry-pi2', 'raspberrypi3']:
         b = BCMRasberryPiBootloader(conffile)
         if not b.configure(old, new):
             log.error("Could not configure bootloader.")
             return False
+    elif runningDevice(conffile) == 'intel-nuc':
```
Page-: Can you store `runningDevice(conffile)` in a var, rather than calling it multiple times

Dockerfile.intel-nuc
```diff
@@ -0,0 +1,9 @@
+FROM resin/nuc-debian:jessie
+
+# Install the dependencies
+RUN apt-get update && apt-get install vim btrfs-tools mtools dosfstools python3 python3-requests python3-sh python3-parted python3-binaryornot jq -y && rm -rf /var/lib/apt/lists/*
+
+# Add the current directoy in the container
+ADD . /app
```
Page-: This should probably be COPY, unless you're using any of the special features of ADD (eg automatic archive decompression)

petrosagg: NIT: we generally prefer `COPY` over `ADD`

petrosagg:
SGTM

Dockerfile.intel-nuc
agherzan: Done.

Page-:
SGTM

PR #17, opened 2016-05-03
Add support for Beaglebone Black

agherzan:

@petrosagg @floion @telphan

### DIFF
--- a/modules/bootconf.py
+++ b/modules/bootconf.py
@@ -180,7 +180,7 @@ class UBootBeagleboneBootloader(BootloaderConfigurator):
                 lines.append(line)
             # If no finduuid found append one
             if fixFindUUID:
-                line.append('finduuid=part uuid mmc ${bootpart} uuid\n')
+                lines.append('finduuid=part uuid mmc ${bootpart} uuid\n')
         with open(uEnvPath + ".tmp", "w") as f:
             for line in lines:
                 f.write(line)
--- a/modules/bootconf.py
+++ b/modules/bootconf.py
@@ -10,6 +10,7 @@
 
 import logging
 import tempfile
+import os
 from .util import *
 
 def configureBootloader(old, new, conffile):
@@ -40,6 +41,8 @@ class BootloaderConfigurator(object):
         self.conf = conf
 
     def applyTextTransformation(self, configurationFile, old, new):
+        log.debug("applyTextTransformation: Apply text replace from " + old + " to " + new + " .")
+
         if not os.path.isfile(configurationFile):
             log.error("applyTextTransformation: configurationFile %s doesn't exist." % configurationFile)
             return False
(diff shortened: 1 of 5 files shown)


modules/bootconf.py
```diff
@@ -118,3 +127,94 @@ def configure(self, old, new):
             return False
 
         return True
+
+class UBootBeagleboneBootloader(BootloaderConfigurator):
+
+    def switchUEnv(self, old, new, uEnvPath):
+        ''' Switch boot partition using uEnv bootpart variable '''
```
petrosagg: Does that mean that we have dual boot partitions as well as dual root partitions?

modules/bootconf.py
```diff
@@ -118,3 +127,94 @@ def configure(self, old, new):
...
+            # If no finduuid found append one
+            if fixFindUUID:
+                line.append('finduuid=part uuid mmc ${bootpart} uuid\n')
+        with open(uEnvPath + ".tmp", "w") as f:
+            for line in lines:
+                f.write(line)
+            # Dump to disk - no need to fsync the parent directory
+            # As we are editing a file - not creating one
+            os.fsync(f)
+        os.rename(uEnvPath + ".tmp", uEnvPath)
```
petrosagg: An `fsync()` on the directory fd is needed after rename to force the rename to be flushed to disk

modules/bootconf.py
```diff
@@ -118,3 +127,94 @@ def configure(self, old, new):
...
+        if not isMounted(bootmountpoint):
+            log.error("transformUEnv: " + bootmountpoint + " is not a mountpoint.")
+            return False
+        uEnvPath = os.path.join(bootmountpoint, 'uEnv.txt')
+        if not os.path.isfile(uEnvPath):
+            log.error("transformUEnv: uEnv.txt seem not to exist in boot partition")
+            return False
+
+        lines=[]
+        fixFindUUID = True
```
petrosagg: Is this specific to a board?

modules/bootconf.py
agherzan: Hmm... the comment is misleading. It actually was meant to sounds as "switch the partition from which to boot - aka root partition). I will change to clear this up.

modules/bootconf.py
agherzan: Sadly it is. This is because it greatly depends on how the default environment is created in uboot. And in this case finduuid is defined in ti_armv7_common.h which is TI specific for armv7 boads.

modules/bootconf.py
agherzan: Correct.

modules/bootconf.py
```diff
@@ -118,3 +134,99 @@ def configure(self, old, new):
...
+                    continue
+                # Tweak finduuid to use a configurable partition
+                if line.startswith('finduuid='):
+                    line = finduuid
+                    fixFindUUID = False
+
+                lines.append(line)
+            # If no finduuid found append one
+            if fixFindUUID:
+                line.append('finduuid=part uuid mmc ${bootpart} uuid\n')
```
petrosagg: `line` -> `lines`

petrosagg:
sgtm

PR #27, opened 2016-07-13
Boot fingerprints and custom kernel modules

agherzan:
Connects to #26
Connects to #24

@petrosagg @floion @telphan @lorenzo-stoakes


modules/fingerprint/tests/testRun/root_tree/file1
```diff
@@ -0,0 +1 @@
+
```
floion: what are these files used for?

modules/fingerprint.py
```diff
@@ -48,47 +51,71 @@ def scan(self):
...
+        # Compute fingerprints file for rootfs partition
+        root_FingerPrintFile = getConfigurationItem(self.conf, "FingerPrintScanner", "root_defaultFingerPrintFile")
+        root_FingerPrintFile = os.path.join(self.root, root_FingerPrintFile)
+
+        # Validate root partition fingerprints
+        if os.path.isfile(root_FingerPrintFile):
+            toReturn = self.do_validateFingerPrints(root_FingerPrintFile, self.root_fingerprints)
         else:
-            # Host OS fingerprint not present on filesystem
+            # Rootfs fingerprints file not present on filesystem
```
floion: shouldn't you error out if no root fingerprint file is found?

agherzan: If there is no such file, we try to match an existing one from the set to ship with the tool. Check the fllow up code and https://github.com/resin-os/resinhup/tree/master/modules/fingerprint/known-images

modules/fingerprint/tests/testRun/root_tree/file1
agherzan: These are helping files for the testcases: https://github.com/resin-os/resinhup/blob/master/modules/fingerprint.py#L120

modules/fingerprint.py
floion: Should there be such a set for the boot fingerprints file also?

agherzan: It should indeed but currently it is not part of this PR. It will come soon.

modules/resinkernel.py
```diff
@@ -0,0 +1,95 @@
...
+    for entry in modinfo:
+        attribute = entry.split(':', 1)[0].strip()
+        value = entry.split(':', 1)[1].strip()
+        if attribute == attr:
+            return value
+
+    return None
+
+class ResinKernel(object):
+    def CustomLoadedModules(self):
```
petrosagg: NIT: why is this method capitalised?

petrosagg:
SGTm

modules/resinkernel.py
agherzan: I use CamelCase for class names as it is recommended in PEP8. Now I do agree that they don't say if the name should start or not with a capital letter but CamelCase can be either). My preference was to start with capital letters for class names as I use CamelCase for methods too (which I do agree that it shouldn't be).

petrosagg: They do specify it. Class names should be CapWords[1], which is camelCase but with the first letter capitalised[2]. Methods should be lowercase separated by underscore[3].

[1] https://www.python.org/dev/peps/pep-0008/#class-names
[2] https://en.wikipedia.org/wiki/CamelCase#Variations_and_synonyms
[2] https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

agherzan: Makes sense. I will change the method name anyway.

modules/fingerprint/tests/testRun/resinhup.conf
```diff
@@ -1,5 +1,9 @@
 [FingerPrintScanner]
-defaultFingerPrintFile: /resinOS-fingerprints
-whitelist:
-     ./fingerprint/tests/testRun/tree/dir3
-     ./fingerprint/tests/testRun/tree/dir2/file3
+
+root_defaultFingerPrintFile: resin-root.fingerprint
+root_whitelist:
+    ./fingerprint/tests/testRun/tree/dir3
```
floion: Shouldn't this be: fingerprint/tests/testRun/root_tree/dir3 or something like that ?

modules/fingerprint/tests/testRun/resinhup.conf
```diff
@@ -1,5 +1,9 @@
...
+    ./fingerprint/tests/testRun/tree/dir2/file3
```
floion: Shouldn't this be: fingerprint/tests/testRun/root_tree/dir2/file3 or something like that ?

conf/resinhup.conf
```diff
@@ -54,6 +55,7 @@ to_keep_files:
 ; persistent over updates.
 ; Only relative paths to boot partition.
 ignore_files:
```
floion: ignore_files section seems to be the same as the boot_whitelist. Why this duplication?

resinhup.py
```diff
@@ -132,7 +132,7 @@ def main():
         images_fingerprint_path = os.path.join(os.path.dirname(sys.modules[__name__].__file__), "modules/fingerprint/known-images")
 
         # Make sure the boot partition device is mounted
-        bootdevice = getBootPartition(self.conf)
+        bootdevice = getBootPartition(args.conf)
```
floion: Can this commit be incorporated in the 5th commit where this change is introduced? Since this is PR level, we can have clean commits merged in the git tree rather than introduce fixes commits like this one

modules/fingerprint.py
```diff
@@ -114,7 +115,8 @@ def validateFingerPrints(self):
 
         # Validate root partition fingerprints
         if os.path.isfile(root_FingerPrintFile):
-            toReturn = self.do_validateFingerPrints(root_FingerPrintFile, self.root_fingerprints)
+            if not self.do_validateFingerPrints(root_FingerPrintFile, self.root_fingerprints):
+                toReturn = False
         else:
```
floion: Can this commit be incorporated in the 4th commit where this change is introduced? Since this is PR level, we can have clean commits merged in the git tree rather than introduce fixes commits like this one

resinhup.py
```diff
@@ -124,6 +125,14 @@ def main():
     else:
         log.info("Can't get current HostOS version. Continuing update...")
 
+    # Check for kernel custom modules
+    if ResinKernel().customLoadedModules():
+        return False
+    else:
+        log.info("No custom loaded modules detected.")
```
floion: Maybe make this explicit that it's about kernel modules.

conf/resinhup.conf
agherzan: `ignore_files` is a list of files which we want to persist across updates. Basically we won't copy these files from the update package. `boot_whitelist` is a list used when checking the fingerprints. Even though they include the same list they have different usage and one is an absolute path one is relative. We can definitely merge them in one as I can't see a case where these would be different.

resinhup.py
agherzan: Sure

modules/fingerprint.py
agherzan: Deal.

resinhup.py
agherzan: Deal.

modules/fingerprint/tests/testRun/resinhup.conf
agherzan: They were changed later on.

modules/fingerprint/tests/testRun/resinhup.conf
agherzan: They were changed later on.

floion:
LGTM

PR #29, opened 2016-09-13
resinhup and dockerhub

agherzan:
Connected to #28

### DIFF
--- a/conf/resinhup.conf
+++ b/conf/resinhup.conf
@@ -51,7 +51,9 @@ to_keep_files:
     /var/lib/connman
 
 [fetcher]
-remote: https://s3.amazonaws.com/resin-production-hup/
+type: dockerhub
+; in the case of a dockerhub fetcher, the remote should be the dockerhub repository path
+remote: resin/resinos
 workspace: /tmp/resinhup
 ; these will be considered fingerprints of the update package to make sure you don't use a mistaken one
 ; relative paths to root of the update package
--- a/conf/resinhup.conf
+++ b/conf/resinhup.conf
@@ -24,6 +24,7 @@ root_whitelist:
     var/log
     etc/wpa_supplicant/entropy.bin
     .rce/key.json
+    usr/bin/run-resinhup.sh
 boot_defaultFingerPrintFile: resin-boot.fingerprint
 boot_whitelist:
     cmdline.txt
--- a/conf/resinhup.conf
+++ b/conf/resinhup.conf
@@ -32,8 +32,6 @@ boot_whitelist:
     EFI/BOOT/grub.cfg
 
 [config.json]
-; Where the conf partition is usually mounted - used on old devices if detected
-default_mountpoint: /mnt/boot
 ; The updater needs the json so it will lookup for it in these locations
 possible_locations: /mnt/boot /mnt/conf /mnt/data-disk
 
--- a/conf/resinhup.conf
+++ b/conf/resinhup.conf
@@ -53,7 +53,9 @@ to_keep_files:
 [fetcher]
 remote: https://s3.amazonaws.com/resin-production-hup/
 workspace: /tmp/resinhup
-update_file_fingerprints: ./resin-boot ./resin-root
+; these will be considered fingerprints of the update package to make sure you don't use a mistaken one
+; relative paths to root of the update package
+update_file_fingerprints: resin-boot/resin-boot.fingerprint resin-root.fingerprint
 
 [staging]
 apiEndpoint: https://api.resinstaging.io
--- a/conf/resinhup.conf
+++ b/conf/resinhup.conf
@@ -33,9 +33,9 @@ boot_whitelist:
 
 [config.json]
 ; Where the conf partition is usually mounted - used on old devices if detected
-default_mountpoint: /mnt/conf
+default_mountpoint: /mnt/boot
 ; The updater needs the json so it will lookup for it in these locations
-possible_locations: /mnt/conf /mnt/data-disk
+possible_locations: /mnt/boot /mnt/conf /mnt/data-disk
 
 [rootfs]
 ; These are files that we want to copy from old rootfs to the new / updated one
--- a/fetcher/dockerhub.py
+++ b/fetcher/dockerhub.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+
+#
+# ** License **
+#
+# Home: http://resin.io
+#
+# Author: Andrei Gherzan <andrei@resin.io>
+#
+
+import os
+import logging
+from docker import Client
+from fetcher.tar import tarFetcher
+from modules.util import *
+
+log = logging.getLogger(__name__)
+
+class dockerhubFetcher(tarFetcher):
+
+    def __init__ (self, conffile, version, remote):
+        super().__init__(conffile, version, remote)
+        machine = runningDevice(conffile)
+        self.remotefile = os.path.join(self.remote + ":" + version + "-" + machine)
+        self.workspacefile = os.path.join(self.workspace, "resinhup.tar")
+
+    def download(self):
+        self.cleanworkspace()
+
+        container_name='resinhup'
+
+        try:
+            cli = Client(base_url='unix://var/run/docker.sock', version='auto')
+        except:
+            log.error("Can't connect to docker daemon.")
+            return False
+
+        # Pull docker image
+        try:
+            log.info("Docker image pull started... this can take a couple of minutes...")
+            log.debug("Pulling " + self.remotefile + " ...")
+            cli.pull(self.remotefile, stream=False)
+        except:
+            log.error("Can't pull update image.")
+            return False
+
+        # Make sure there is no 'resinhup' container
+        try:
+            cli.remove_container(container_name, force=True)
+        except:
+            pass
+
+        # Create resinhup container
+        try:
+            cmd = '/bin/bash'
+            container = cli.create_container(image=self.remotefile, command=cmd, name=container_name)
+        except:
+            log.error("Can't create temporary update container.")
+            return False
+
+        # Export the temporary resinhup container as tar archive
+        try:
+            strm = cli.export(container=container.get('Id'))
+        except:
+            log.error("Can't export tar archive update file.")
+            return False
+        with open(self.workspacefile, 'wb') as fd:
+            for chunk in strm.stream(1000000):
+                fd.write(chunk)
+
+        return True
--- a/fetcher/dockerhub.py
+++ b/fetcher/dockerhub.py
@@ -8,8 +8,8 @@
 # Author: Andrei Gherzan <andrei@resin.io>
 #
 
-import os
 import logging
+import os
 from docker import Client
 from fetcher.tar import tarFetcher
 from modules.util import *
--- a/fetcher/dockerhub.py
+++ b/fetcher/dockerhub.py
@@ -22,7 +22,6 @@ class dockerhubFetcher(tarFetcher):
         super().__init__(conffile, version, remote)
         machine = runningDevice(conffile)
         self.remotefile = os.path.join(self.remote + ":" + version + "-" + machine)
-        self.workspacefile = os.path.join(self.workspace, "resinhup.tar")
 
     def download(self):
         self.cleanworkspace()
@@ -64,8 +63,6 @@ class dockerhubFetcher(tarFetcher):
         except:
             log.error("Can't export tar archive update file.")
             return False
-        with open(self.workspacefile, 'wb') as fd:
-            for chunk in strm.stream(1000000):
-                fd.write(chunk)
+        self.updatefilestream = strm
 
         return True
--- a/fetcher/fetcher.py
+++ b/fetcher/fetcher.py
@@ -9,8 +9,8 @@
 #
 
 import logging
-from fetcher.tar import tarFetcher
 from fetcher.dockerhub import dockerhubFetcher
+from fetcher.tar import tarFetcher
 
 log = logging.getLogger(__name__)
 
--- a/fetcher/fetcher.py
+++ b/fetcher/fetcher.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python3
+
+#
+# ** License **
+#
+# Home: http://resin.io
+#
+# Author: Andrei Gherzan <andrei@resin.io>
+#
+
+import logging
+from fetcher.tar import tarFetcher
+from fetcher.dockerhub import dockerhubFetcher
+
+log = logging.getLogger(__name__)
+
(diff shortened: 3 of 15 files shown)


modules/util.py
```diff
@@ -642,5 +649,19 @@ def testSafeDirCopyFile(self):
         dst = "./modules/util/safedircopy/dir3"
         self.assertFalse(safeDirCopy(src, dst))
 
+    def testSafeDirCopyIgnoreDir(self):
+        src = "./modules/util/safedircopy/dir1"
+        dst = "./modules/util/safedircopy/dir3"
+        self.assertTrue(safeDirCopy(src, dst, ignore=['ignore-dir']))
```
floion: Why not use the safeCopy instead here?

modules/util.py
```diff
@@ -642,5 +649,19 @@ def testSafeDirCopyFile(self):
...
+        self.assertFalse(os.path.isdir(os.path.join(dst, "ignore-dir")))
+        shutil.rmtree(dst) # cleanup
+
+    def testSafeDirCopyIgnoreFile(self):
+        src = "./modules/util/safedircopy/dir1"
+        dst = "./modules/util/safedircopy/dir3"
+        self.assertTrue(safeDirCopy(src, dst, ignore=['ignore-file']))
```
floion: Why not use safeCopy instead here?

conf/resinhup.conf
```diff
@@ -53,7 +53,9 @@ to_keep_files:
 [fetcher]
```
floion: Can you rephrase the commit log please? :)

modules/util.py
agherzan: It would have worked with that too. I will change. Thanks.

modules/util.py
agherzan: It would have worked with that too. I will change. Thanks.

conf/resinhup.conf
agherzan: Will try. :)

fetcher/dockerhub.py
```diff
@@ -0,0 +1,88 @@
...
+import logging
+from docker import Client
+from fetcher.tar import tarFetcher
+from modules.util import *
+
+log = logging.getLogger(__name__)
+
+class dockerhubFetcher(tarFetcher):
+
+    def __init__ (self, conffile, version, remote):
```
floion: Why don't you call the parent's class __init__ and override only what you need?

fetcher/fetcher.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
```
floion: Why not use this class to move the bulk from the tarFetcher class here?

fetcher/dockerhub.py
agherzan: I wanted to change this at the end but then I was to lazy to do it. I will have it. Given that this state was tested as it is, would it be fine to have a top commit for this?

fetcher/fetcher.py
agherzan: Because the methods in the tar fetcher are tighted to tar archiving. We can see what we can split out but I propose to do it as a separate PR. Cause that is more like a refactoring commit.

fetcher/dockerhub.py
floion: Either way is fine

fetcher/fetcher.py
floion: As you consider best

fetcher/dockerhub.py
agherzan: Doing.

conf/resinhup.conf
floion: "The resinhup update package have changes the format" - this still throws me off

scripts/docker-build-and-push.sh
```diff
@@ -77,7 +77,7 @@ for dockerfile in $DOCKERFILES; do
     fi
     printf "${GREEN}Running build for $device using $dockerfile ...${NC}\n"
     docker build -t resinhup-$device:$TAG -f ../$dockerfile $SCRIPTPATH/..
-    printf "${GREEN}Tag an push for $device ...${NC}\n"
+    printf "${GREEN}Tag and push for $device ...${NC}\n"
```
floion: Typo in the commit log also: "sypo" there

agherzan: Fixed

conf/resinhup.conf
agherzan: :) Fixed.

floion:
LGTM

fetcher/dockerhub.py
```diff
@@ -0,0 +1,80 @@
...
+        super().__init__()
+        self.remotefile = os.path.join(self.remote + ":resinhup-" + machine + "-" + version)
```
petrosagg: As commented here https://github.com/resin-os/resin-yocto-scripts/pull/33#r79108514 , let's change the tag to be without `resinhup-`

fetcher/dockerhub.py
```diff
@@ -0,0 +1,80 @@
...
+            cli.remove_container(container_name, force=True)
+        except:
+            pass
+
+        # Create resinhup container
+        try:
+            # Docker mounts a set of files at container creation
+            # https://docs.docker.com/engine/userguide/networking/default_network/configure-dns/
+            # We still need all these files as they are important in the update process
+            cmd='bash -c "umount /etc/resolv.conf ; umount /etc/hostname ; umount /etc/hosts"'
```
petrosagg: This has no effect on the export

fetcher/dockerhub.py
```diff
@@ -0,0 +1,80 @@
...
+            container = cli.create_container(image=self.remotefile, command=cmd, name=container_name)
+        except:
+            log.error("Can't create temporary update container.")
+            return False
+
+        # Run container
+        try:
+            response = cli.start(container=container.get('Id'), privileged=True)
```
petrosagg: No need to run any commands or start the container. You can export right after creating

fetcher/dockerhub.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # Export the temporary resinhup container as tar archive
+        try:
+            strm = cli.export(container=container.get('Id'))
+        except:
+            log.error("Can't export tar archive update file.")
+            return False
+        with open(self.workspacefile, 'wb') as fd:
+            for chunk in strm.stream(1000000):
+                fd.write(chunk)
```
petrosagg: Adhering to how the original tar fetcher works means that now we need double space in the data partition. This would be much better if it passed the stream to `tarfile` directly

fetcher/tar.py
```diff
@@ -99,18 +101,26 @@ def unpack(self, downloadFirst=False):
         update = tarfile.open(name=self.workspacefile, mode='r:*')
```
petrosagg: Instead of passing a file here you can replace with a file object. Here is how you would open the stream returned by dockerpy:

```python
import docker
import tarfile

cli = docker.Client()
ctnr = cli.create_container('busybox', 'true')
strm = cli.export(ctnr)
tar = tarfile.open(mode='r|*', fileobj=strm)
tar.list()
```

The important bit is the `r|*` (instead of `r:*`) mode which tells tarfile to not perform any `seek()` operations and therefore can operate on a stream

fetcher/dockerhub.py
```diff
@@ -0,0 +1,80 @@
...
+        # Run container
+        try:
+            response = cli.start(container=container.get('Id'), privileged=True)
+        except:
+            log.error("Can't start temporary update container.")
+            return False
+
+        # Export the temporary resinhup container as tar archive
+        try:
+            strm = cli.export(container=container.get('Id'))
```
petrosagg: You can pass the `container` variable directly

fetcher/tar.py
```diff
@@ -99,18 +101,26 @@ def unpack(self, downloadFirst=False):
...
         update.extractall(self.workspaceunpack)
```
petrosagg: Why is `workspaceunpack` not the unused root partition? This means that the current PR needs 3*(root_size) of free space available in the data partition.

fetcher/tar.py
```diff
@@ -99,18 +101,26 @@ def unpack(self, downloadFirst=False):
...
-        rootfsfiles = os.listdir(self.workspace + "/update/resin-root")
-        self.rootfsarchive = self.workspace + "/update/resin-root/" + os.listdir(self.workspace + "/update/resin-root")[0]
-
         log.debug("Unpacked " + self.workspacefile + " in " + self.workspaceunpack)
         return True
 
+    def unpackQuirks(self, location):
+        '''Copies the content of <quirks> directory from an update to <location>.'''
+        quirks_path = os.path.join(self.workspaceunpack, 'quirks')
+        log.info("Unpacking rootfs quirks from " + quirks_path + " to " + self.workspaceunpack + ".")
```
petrosagg: According to the code, unpacking happens from `quirks_path` to `location`, not `self.workspaceunpack`

fetcher/tar.py
```diff
@@ -99,18 +101,26 @@ def unpack(self, downloadFirst=False):
...
+            return False
+        log.info("Unpack rootfs quirks done.")
+        return True
+
     def unpackRootfs(self, location):
         log.info("Unpack rootfs started... this can take a couple of seconds or even minutes...")
-        update = tarfile.open(self.rootfsarchive)
-        update.extractall(location)
-        log.debug("Unpacked rootfs " + self.rootfsarchive + " in " + location)
+        safeCopy(self.workspaceunpack, location, sync=False, ignore=['resin-boot', 'quirks'])
```
petrosagg: Why don't need the ignore mechanism at all if the operations happen in the right order. Can we only unpack __once__ into the inactive root partition and then do the two copy operations for `resin-boot` and `quirks` from there?

fetcher/tar.py
floion: Nice. This would also mean some small changes in the tarFetcher class to define a boolean that would signal if we get the stream from the dockerhubFetcher or prepare it ourselves here in the unpack() method.

fetcher/dockerhub.py
agherzan: Deal.

fetcher/dockerhub.py
agherzan: This was an old implementation for the fix of hostname and co shadowing. I forgot to remove it. Good catch.

fetcher/dockerhub.py
agherzan: This was an old implementation for the fix of hostname and co shadowing. I forgot to remove it. Good catch.

fetcher/dockerhub.py
agherzan: The container variable holds a dict. While the export expects a string as the container argument `container (str): The container to export`. Did you mean `container_name`?

fetcher/dockerhub.py
agherzan: That makes a lot of sense.

fetcher/tar.py
agherzan: @petrosagg Definitely a very good point.
@floion No. Actually all we need is to replace workspacefile by IO stream.

fetcher/tar.py
agherzan: Type. Fixing.

fetcher/tar.py
agherzan: See comment above about rootfs size.

fetcher/dockerhub.py
floion: As Petros pointed out in a comment in fetcher/tar.py, the following piece of code does the export using the variable returned by create_container:

import docker
import tarfile

cli = docker.Client()
ctnr = cli.create_container('busybox', 'true')
strm = cli.export(ctnr)
tar = tarfile.open(mode='r|*', fileobj=strm)
tar.list()

fetcher/tar.py
floion: Well, if you also want to keep the tarFetcher, you will need to adapt it to work with streams I think to better accommodate both cases. So it can work with such a boolean, plus creating a stream out of the tar archive the tar fetcher downloads

fetcher/tar.py
floion: Why not? Isn't the unused rootfs formatted by resinhup at the beginning of the process?

agherzan: @floion the resinhup update packages (the archive) containes rootfs + boot + some
files. Extracting this in rootfs will most probably go over the rootfs size.

fetcher/tar.py
agherzan: Indeed changes are needed but I don't need a flag for this. I can simply use a stream. Implemented in the last push.

fetcher/dockerhub.py
agherzan: It is just an undocumented way of using it. It is handled by code https://github.com/docker/docker-py/blob/c5a81002c6af649ed5deef44b154b05627559e1b/docker/utils/decorators.py#L7 but the documentation clearly states that it needs a str as an argument.

fetcher/tar.py
agherzan: There is a problem too in extracting the boot files directly from the tar to the boot partition: handling atomicity which is currently handled as a copy method in resinhup.

modules/updater.py
```diff
@@ -141,7 +146,7 @@ def rootfsOverlay(self):
             if not os.path.exists(src_full_path):
                 log.warn(src_full_path + " was not found in your current mounted rootfs. Can't overlay.")
                 continue
-            if not safeCopy(src_full_path, self.tempRootMountpoint + dst):
+            if not safeCopy(src_full_path, self.tempRootMountpoint + dst, sync=False):
```
petrosagg: I don't think this is a necessary change especially since it introduces a subtle way of breaking the updater which is the power cutting after having switched the bootloader but before these writes have been persisted. Since performance isn't an issue I suggest keeping all the operations safe

fetcher/tar.py
```diff
@@ -59,58 +59,58 @@ def download(self):
             log.error("HTTP status code: " + str(r.status_code))
             return False
 
-        with open(self.workspacefile, 'wb') as fd:
-            for chunk in r.iter_content(1000000):
-                fd.write(chunk)
+        self.updatefilestream = StringIO(r.content)
```
petrosagg: This is a really bad idea. It will load the whole tar archive into memory.

fetcher/tar.py
petrosagg: Ok. Understood. This is something to fix in the next releases.

modules/updater.py
agherzan: Got it. Sure.

PR #33, opened 2016-11-15
updater: remove outdated code and misleading error message

michal-mazurek:
btrfs is no longer used for resin-data. This error message is misleading.

Also remove some commented out code.


michal-mazurek:
@agherzan

PR #36, opened 2016-12-01
docker-build-and-push: add configurable registry

michal-mazurek:
The new option is -r, or --registry

Also fix a bug that manifested itself if two images with the same
name, but different tags existed.


michal-mazurek:
@agherzan @floion @telphan @petrosagg

PR #37, opened 2016-12-01
docker-build-and-push.sh: help message tweaks

michal-mazurek:
Make the help message look a bit more like a unix manpage.


michal-mazurek:
@agherzan @floion @telphan @petrosagg

PR #38, opened 2016-12-05
resinhup.py: fail if asked to update to version 2.0.0 or greater

michal-mazurek:


michal-mazurek:
@agherzan @floion @petrosagg

resinhup.py
```diff
@@ -117,6 +117,13 @@ def main():
         log.error("Resinhup is not supported for the resinOS version requested.")
         return False
 
+    # Maximum resinOS version
+    # XXX not supported yet
```
agherzan: Can we have a fixme tag here? I'm not sure what XXX is.

resinhup.py
```diff
@@ -117,6 +117,13 @@ def main():
...
+    max_version = "2.0.0"
+    if StrictVersion(args.version) >= StrictVersion(max_version):
```
agherzan: I don't think this plays wall if you request a version like 2.0.0-beta6.

agherzan:
Needs rebase too.

agherzan:
@michal-mazurek something is wrong with this PR

resinhup.py
michal-mazurek: It's a common marker for "problematic or misguiding code": https://en.wikipedia.org/wiki/Comment_(computer_programming)
I don't think it's quite the same as a FIXME.

resinhup.py
michal-mazurek: Indeed, the program fails: "ValueError: invalid version number '2.0.0-beta6'"

resinhup.py
```diff
@@ -117,6 +117,13 @@ def main():
         log.error("Resinhup is not supported for the resinOS version requested.")
```
michal-mazurek: I think this expression suffers from the same problem as pointed out below ("2.0.0-beta6"). "2.0.0-beta6" is not a valid version string supported by distutils.version which this program uses. One solution is to stop using the non-standard versions. The other would be to sanitize these versions and convert them to something that's supported. Comments?

resinhup.py
agherzan: TIL :+1:

resinhup.py
agherzan: It need a little parsing here in order to avoid it.

resinhup.py
agherzan: I would say that beta version shouldn't be updated. If you are running a beta version you are aware that resinhup will not work for you so you will have to reflash at some point. Now, in order to detect this we can do the following:
1. if beta bail out
2. if force, update even beta.

I'm a little concerned about the force flag too. The peoblem is that if you force the version it will automatically force the fingerprints check too. Maybe we should use additional flags? Or some other smarter mechanism.

michal-mazurek: We don't support 2.X yet, so 2. is irrelevant at present. So I say leave it as it is now. Right now the program fails if the version string is not standard (is beta).

willnewton:
This doesn't seem to be needed, we can revisit if it becomes a problem.

PR #41, opened 2016-12-08
resinhup.conf: add failing 1.8.0 files to the whitelist

michal-mazurek:


michal-mazurek:
@agherzan @floion @petrosagg

PR #43, opened 2016-12-09
rce fixes

agherzan:


Dockerfile.beaglebone-black
```diff
@@ -1,7 +1,12 @@
 FROM resin/beaglebone-debian:jessie
 
 # Install the resinhup dependencies
-RUN apt-get update && apt-get install vim btrfs-tools mtools dosfstools python3 python3-pip python3-urllib3 python3-sh python3-parted python3-binaryornot jq -y && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install wget vim btrfs-tools mtools dosfstools python3 python3-urllib3 python3-sh python3-parted python3-binaryornot jq -y && rm -rf /var/lib/apt/lists/*
+
+# Install pip manually otherwise we will get a broken dependency of docker-py: requests-2.12.2
+# https://github.com/docker/docker-py/issues/1321
+RUN wget https://bootstrap.pypa.io/get-pip.py
```
petrosagg: How is pip related to requests and broken docker-py? I've read the linked issue but I don't get why we can't just do a `pip3 install requests==2.12.3` or something like that

fetcher/dockerhub.py
```diff
@@ -21,6 +22,12 @@ class dockerhubFetcher(tarFetcher):
     def __init__ (self, conffile, version, remote):
         super().__init__(conffile, version, remote)
         machine = runningDevice(conffile)
+
+        docker_version = "1.1.5"
+        if StrictVersion(getCurrentHostOSVersion(conffile)) < StrictVersion(docker_version):
```
petrosagg: This looks like you're comparing a docker version with a host OS version which doesn't make any sense. I got it in the end but `docker_version` needs a better name. How about a `PRE_DOCKER_OS` constant?

Dockerfile.beaglebone-black
agherzan: Installing pip from debian repo brings in requests too. But debian has the broken version while pip has a newer version which fixes the problem. And requests is a dep of the pip package docker py too. So installing dockerpy from pip is enough.

fetcher/dockerhub.py
agherzan: Fixed. Please rereview @petrosagg

PR #44, opened 2016-12-15
Fixes

agherzan:


modules/util.py
```diff
@@ -531,13 +531,9 @@ def safeDirCopy(src, dst, sync=True, ignore=[]):
 
     # Copy each file in the structure of src to dst
     for root, dirs, files in os.walk(src):
-
             # Directories
+            dirs[:] = list(filter(lambda d: not d in ignore, dirs))
```
petrosagg: NIT: this can also be a comprehension: `dirs[:] = [d for d in dirs if not d in ignore]`

floion:
"util.py: Fix bug when altering the varible on which the iteration is …
…done" : can you describe such a case when the d variable is changes, I assume?

floion:
If it's related to the fact that you should not modify the content of a list while looping over it, then please change the commit log and the subsequent CHANGELOG entry to reflect that fact, rather than just saying this fixes the issues when there are more than one directories to be ignored, which is not the real issue.

floion:
Okay, otherwise (besides a more verbose commit log as noted in my previous comments) approved

modules/util.py
agherzan: Addressed.

agherzan:
@floion addressed. Thank you.

PR #46, opened 2016-12-21
Fixes

agherzan:
supervisor.conf - can be modified by update-resin-supervisor
update-resin-supervisor - is scp'd via ssh/vpn while running resinhup ssh wrapper

Fixes https://github.com/resin-os/resinhup/issues/45

### DIFF
--- a/conf/resinhup.conf
+++ b/conf/resinhup.conf
@@ -25,6 +25,8 @@ root_whitelist:
     etc/wpa_supplicant/entropy.bin
     .rce/key.json
     usr/bin/run-resinhup.sh
+    usr/bin/update-resin-supervisor
+    etc/supervisor.conf
 boot_defaultFingerPrintFile: resin-boot.fingerprint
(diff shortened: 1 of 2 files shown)


conf/resinhup.conf
```diff
@@ -25,6 +25,8 @@ root_whitelist:
     etc/wpa_supplicant/entropy.bin
     .rce/key.json
     usr/bin/run-resinhup.sh
+    usr/bin/update-resin-supervisor
```
floion: Isn't /usr/bin/update-resin-supervisor already present in the rootfs? Why is it scp'ed by resinhup ssh wrapper script?

agherzan: Because the update is running this script and needs to be sure it is updated and working properly (for devices where things like version thought arguments were not implemented or for devices with broken script due to various bugs).

agherzan: This will be removed in time.

PR #47, opened 2016-12-27
docker-build-and-push.sh: mention the default --registry argument in the help message

michal-mazurek:


michal-mazurek:
connect https://github.com/resin-os/resinhup/issues/42

scripts/docker-build-and-push.sh
```diff
@@ -21,8 +21,9 @@ Options:
         Build and push only this Dockerfile. Otherwise all found will be used.
 
   -r, --registry
-        The Docker registry to push to, without the trailing slash.
+        List of one or more Docker registries to push to, without the trailing slash.
         Remember to change the corresponding value in conf/resinhup.conf as well.
+        Default: "$REGISTRY".
```
agherzan: This is not actually working as you would expect. If you runt the script with `--registry "foo" --help`, it would show a wrong default value.

michal-mazurek: I didn't think of that. Fixed.

PR #51, opened 2017-01-05
resinhup.py: attempt to retry a fetch if it failed

michal-mazurek:


michal-mazurek:
connect https://github.com/resin-os/resinhup/issues/49

PR #53, opened 2017-01-09
resinhup.conf: whitelist splash/resin-logo.png

michal-mazurek:


michal-mazurek:
connect https://github.com/resin-os/meta-resin/issues/538

PR #56, opened 2017-01-20
Implement resinhup downgrades

agherzan:
After merging this:
1. release 1.1.1
2. use this tag in https://github.com/resin-os/meta-resin/blob/1.X/meta-resin-common/recipes-support/resinhup/resinhup/run-resinhup.sh#L4 and append it to (this)[https://github.com/resin-os/meta-resin/pull/549] PR


resinhup.py
```diff
@@ -47,6 +47,8 @@ def main():
                       help = "Configuration file to be used. Default: " + default_resinhup_conf_file)
     parser.add_argument('-f', '--force', action = 'store_true', dest = 'force', default = False,
```
petrosagg: so `--force` is obsolete now?

fetcher/dockerhub.py
```diff
@@ -34,7 +34,7 @@ def __init__ (self, conffile, version, remote):
     def download(self):
         self.cleanworkspace()
 
-        container_name='resinhup'
+        container_name='resinos'
```
petrosagg: How did this work in the past? Is it overridable by an argument?

resinhup.py
agherzan: No. It is for all the other checks but version (currently the remaining ones are fingerprints and custom modules I think).

fetcher/dockerhub.py
agherzan: No. It was working because there was no other container called resinhup. Meta-resin PR now uses this name.

PR #66, opened 2017-05-30
WIP: Alpine-based resinhup images

imrehg:
Trying to save some space in the resinHUP image, so the image download is faster on the device, and trying Alpine for that.

Changes so far:

* using Alpine Edge (because dependencies are in there that are needed at the moment)
* time-tagged base image
* finding the packages explicitly which supply each of the tools that are used in the scripts
* installing Python dependencies with pip instead form the system
* combining install step so makedepends can be removed at the end, saving considerable space

The images seem to be from ~100MB -> ~57MB. Tested on RPi and RPi3 so far. 

Work in progress, opening PR for feedback! Cheers! :)


willnewton:
This looks like a useful change. Would it be possible to use a non-edge version of Alpine?

imrehg:
@willnewton great point, yeah, hoping to do that too, I think 3.6 should work, except that our alpine images do not come in that flavor yet (see for example [raspberrypi3-alpine](https://hub.docker.com/r/resin/raspberrypi3-alpine/tags/)). Filing an issue for that. :)

In the meantime I've tested the BBB update, and updating both from 1.8 and 1.24 (that are currently downloadable, and larger change between them), both seems okay. Outstanding is NUC and pre-Docker-1.10 release updates, for completeness

Hunting down the source of some new DEBUG messages around `safeFileCopy`, shouldn't be an issue, but would be useful know what tool behaves differently on Alpine than on Debian, things like this:

```
[...snip...]
WARNING : safeFileCopy: Destination file /tmp/resinhup-g5sjchc2/overlays/pps-gpio-overlay.dtb already exists. Will overwrite.
DEBUG : Copied /tmp/resinhup/update/resin-boot/overlays/pps-gpio-overlay.dtb to /tmp/resinhup-g5sjchc2/overlays/pps-gpio-overlay.dtb
DEBUG : is_binary: '/tmp/resinhup/update/resin-boot/overlays/rpi-ft5406-overlay.dtb'
DEBUG : nontext_ratio1: 0.518595041322314
DEBUG : nontext_ratio2: 0.981404958677686
DEBUG : is_likely_binary: False
DEBUG : SHIFT_JIS Japanese prober hit error at byte 2
DEBUG : EUC-JP Japanese prober hit error at byte 1
DEBUG : GB2312 Chinese prober hit error at byte 1
DEBUG : EUC-KR Korean prober hit error at byte 1
DEBUG : CP949 Korean prober hit error at byte 1
DEBUG : Big5 Chinese prober hit error at byte 1
DEBUG : EUC-TW Taiwan prober hit error at byte 1
DEBUG : windows-1251 Russian confidence = 0.0
DEBUG : KOI8-R Russian confidence = 0.0
DEBUG : ISO-8859-5 Russian confidence = 0.0
DEBUG : MacCyrillic Russian confidence = 0.0
DEBUG : IBM866 Russian confidence = 0.01
DEBUG : IBM855 Russian confidence = 0.0
DEBUG : ISO-8859-7 Greek confidence = 0.7630861646373661
DEBUG : windows-1253 Greek confidence = 0.7630861646373661
DEBUG : ISO-8859-5 Bulgairan confidence = 0.0
DEBUG : windows-1251 Bulgarian confidence = 0.0
DEBUG : TIS-620 Thai confidence = 0.0
DEBUG : ISO-8859-9 Turkish confidence = 0.5621562063450571
DEBUG : windows-1255 Hebrew confidence = 0.0
DEBUG : windows-1255 Hebrew confidence = 0.0
DEBUG : windows-1255 Hebrew confidence = 0.0
DEBUG : windows-1251 Russian confidence = 0.0
DEBUG : KOI8-R Russian confidence = 0.0
DEBUG : ISO-8859-5 Russian confidence = 0.0
DEBUG : MacCyrillic Russian confidence = 0.0
DEBUG : IBM866 Russian confidence = 0.01
DEBUG : IBM855 Russian confidence = 0.0
DEBUG : ISO-8859-7 Greek confidence = 0.7630861646373661
DEBUG : windows-1253 Greek confidence = 0.7630861646373661
DEBUG : ISO-8859-5 Bulgairan confidence = 0.0
DEBUG : windows-1251 Bulgarian confidence = 0.0
DEBUG : TIS-620 Thai confidence = 0.0
DEBUG : ISO-8859-9 Turkish confidence = 0.5621562063450571
DEBUG : windows-1255 Hebrew confidence = 0.0
DEBUG : windows-1255 Hebrew confidence = 0.0
DEBUG : windows-1255 Hebrew confidence = 0.0
DEBUG : detected_encoding: {'encoding': 'ISO-8859-7', 'confidence': 0.7630861646373661, 'language': 'Greek'}
DEBUG : failure: decodable_as_unicode: False
DEBUG : has nulls:True
```

imrehg:
@agherzan @willnewton rewritten this PR with 3 main fronts:

* use 3.6 alpine images (no edge anymore) as per comments, now that we have those
* using multi-stage build in Docker (size should be the same as hand-optimized image, but it is cleaner this way, having 2 steps - one python build, one dependencies and copy-over). The size now is ~45MB (from ~95MB)
* moving all app files into an `app` folder, so won't accidentally include extra files from the folder in the container due to `COPY . /app` (as it happened with me while working on it and having work in progress and debug stuff in the root folder of the project). Let me know if this is acceptable, or should be done differently.

One successful RPI3 HUp was done, log attached: [resinhup.log.zip](https://github.com/resin-os/resinhup/files/1105068/resinhup.log.zip)

imrehg:
ping @agherzan and @willnewton for feedback

willnewton:
Did you figure out the debug messages you mentioned above?

Dockerfile.beaglebone-black
```diff
@@ -1,18 +1,43 @@
...
+#############
+FROM resin/beaglebone-black-alpine:3.6-20170623 AS build
 
-# Install the resinhup dependencies
-RUN apt-get update && apt-get install wget vim btrfs-tools mtools dosfstools python3 python3-urllib3 python3-sh python3-parted python3-binaryornot jq -y && rm -rf /var/lib/apt/lists/*
+# Install build requirements, create modules, then remove superfluous requirements
+## Note:
+# apk add is two lines: first is runtime dependencies, second is build time dependencies that we can remove after install
+RUN apk add --no-cache \
+      'python3<3.7' gcc libc-dev parted-dev python3-dev
```
willnewton: Is it worth documenting why we require that version of python?

willnewton:
It would also be good to run a test on original Raspberry Pi (ARMv6) to make sure it still works. Otherwise this looks like a good change to me.

imrehg:
@willnewton I think the debug message is from "is_binary" python package, in this line: https://github.com/resin-os/resinhup/blob/master/modules/util.py#L263-L264

Going to run the Alpine-based images on all the supported types before taking off the "WIP" from this PR

Dockerfile.beaglebone-black
imrehg: Removed that requirement, as I thought it was needed for the later COPY not to fail silently, but actually it does not fail silently, if python version changes, the build would bail out at the COPY step. Removed this version requirement!

imrehg:
@willnewton @agherzan: run this alpine-based resinhup on RPi/RPi2/RPi2/NUC/BBB (1.8.0 -> 1.26.0) and seems to work well.

Dockerfile.beaglebone-black
```diff
@@ -1,18 +1,41 @@
...
+
+# Required packages
+# blkid: blkid
+
+# dosfstools: dosfslabel
+# e2fsprogs-extra: e2label
+# kmod: lsmod, rmmod
+# util-linux: lsblk
+RUN apk add --no-cache \
+      blkid btrfs-progs btrfs-progs-extra dosfstools e2fsprogs-extra jq kmod mtools parted python3 util-linux vim wget \
```
agherzan: I think we can get rid of vim. Most probably is there for debugging.

Dockerfile.beaglebone-black
```diff
@@ -1,18 +1,41 @@
...
+###############
+# Shipped image
+###############
+FROM resin/beaglebone-black-alpine:3.6
+
+WORKDIR /app
+
+# Required packages
+# blkid: blkid
+
```
agherzan: Let's remove this new line.

Dockerfile.beaglebone-black
imrehg: @agherzan ack

Dockerfile.beaglebone-black
imrehg: Ack

imrehg:
@agherzan 

1. As I see, at least ~6MB uncompressed (base image is different, so cannot check the compressed), mainly because currently `.git` is copied in there too due to the wildcard copy. This change is mostly for clarity and maintainability, howerver. There are three kinds of files here: 1. those needed by docker to build the image, 2. the actual app running in resinhup, 3. helper scripts etc. Just trying to make sure that the 2. and 3. types are clearly separated. I can do a separate PR for this part of the chane too if needed, or can omit it you so prefer at all.
2. definitely, that's why it's marked a "minor" even (imho justified to be above patch level), which would bump to `v1.2.0` if I understand versionist properly

Thanks for the feedback :)

agherzan:
1. uh yeah - I forgot about the .git directory - that makes sense now
2. I obviously forgot that we added revisionist support - looks good.

PR #69, opened 2017-06-08
legacy-config: fix bug when migrating config.json

imrehg:
Trying to copy with wrong syntax (to a directory, instead of a file),
which breaks update process.

Commit-type: patch
Signed-off-by: Gergely Imreh <imrehg@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file.
 
+ - legacy-config: fix copy bug when migrating config.json
+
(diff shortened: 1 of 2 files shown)


willnewton:
Needs a changelog entry? Otherwise looks good to me.

imrehg:
Yeah, adding that, cheers! :)

imrehg:
Tested with 1.0.0-pre / 0.0.14

imrehg:
Yeah, because you wrote it :)

Issue #70, opened 2017-06-08
Add versionist/versionbot support


imrehg:
So no explicit changelog entry changes need to be made, but all that's automatic.

willnewton:
@imrehg can this be closed now?

PR #71, opened 2017-06-08
versionist: add support for versionist generation of CHANGELOG.md

imrehg:
This is the first step to towards automatically generate CHANGELOG with versionbot later, and take off some of the burden of the developers, have cleaner PRs. Versionist/versionbot also widely used around resin, so helps sharing the tooling.

* As is, if merged, run [versionist](https://github.com/resin-io/versionist) in the root of the repo, and it calculates the new version number based on the `Change-type` entry in any of the commits since the last version, and adds the entry to the CHANGELOG. In this manual case, the person releasing has to commit the Changelog changes in the release commit (and update the tag), but the PR writer does not need to add any entries, just good first line description in the commit for what the change is.
* In the next step, versionbot can be enabled, and then merging and tagging (manual above) can be done by applying labels, and the bot doing it automagically.

Connects-to: #70 
Change-type: patch
---- Autogenerated Waffleboard Connection: Connects to #70


agherzan:
One problem here: we need to update https://github.com/resin-os/resinhup/blob/master/meta/resinhupmeta.py#L14 too

imrehg:
@agherzan added support for `resinhupmeta.py` as well, let's see if `sed` is a workable path :P

agherzan:
Looks good as an workaround for me. @hedss , thoughts? BTW, can we integrate with revisionist as a first stage in such a way that automatic releases per PR will be deactivated? I'm interested for now in `Patch-type` check and I'll do the release by running versionist manually. This as well will give me the chance to build and deploy the images too (we should create jenkins jobs for this on every tag and every master change - tag latest).

imrehg:
@agherzan is that the first step in the PR? if merged, you can run `versionist` locally, and then it would do the updates (but not tagging!) as versionbot would do. Would that be what you mean?

Versionbot is not activated on this repo yet, I think, so it's all manual anyways. 

What do you mean by `revisionist`, though? :) And it's `Change-type:`, not `Patch-type:`

agherzan:
Fixed my message above @imrehg .

Indeed I want to keep the release manual for now: versionist + commit + tag + push.

imrehg:
Updated PR due to project layout changes in #66

PR #75, opened 2017-07-03
Add a script to upgrade resinOS 1.x to 2.x

willnewton:
This script should be able to upgrade a Raspberry Pi or Beaglebone device from resinOS 1.16 or above to a suitable version of resinOS 2.x. It still requires a small tweak to update the supervisor after an upgrade but resinOS 2.0.8 should fix this.


upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,409 @@
...
+esac
+root_dev=${root_part%p2}
+
+# Check if we need to install some more extra tools
+if ! version_gt $VERSION $PREFERRED_HOSTOS_VERSION && ! [ "$VERSION" == $PREFERRED_HOSTOS_VERSION ]; then
+    log "Host OS version $VERSION is less than $PREFERRED_HOSTOS_VERSION, installing tools..."
+    case $BINARY_TYPE in
+	arm)
+	    # Remove symlink to busybox tar
+	    rm /bin/tar
```
agherzan: You can avoid this by using a separate path in PATH. And maybe it would be clearer to keep the binaries in a specific location and just prepend that location in PATH.

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,409 @@
...
+	    # XXX move these files somewhere else
+	    curl -s -L -o /bin/tar https://github.com/resin-os/resinhup/raw/onetotwo/upgrade-binaries/arm/tar
```
agherzan: I would replace `https://github.com/resin-os/resinhup/raw/onetotwo/upgrade-binaries` by a variable so when we move them we can make the change easier.

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,410 @@
...
+	arm)
+	    download_uri=https://github.com/resin-os/resinhup/raw/onetotwo/upgrade-binaries/$BINARY_TYPE
+	    curl -s -L -o $tools_path/tar $download_uri/tar
+	    chmod 755 $tools_path/tar
+	    curl -s -L -o $tools_path/e2label $download_uri/e2label
+	    chmod 755 $tools_path/e2label
+	    curl -s -L -o $tools_path/mkfs.ext4 $download_uri/mkfs.ext4
+	    chmod 755 $tools_path/mkfs.ext4
+	    curl -s -L -o $tools_path/resize2fs $download_uri/resize2fs
+	    chmod 755 $tools_path/resize2fs
```
imrehg: Would it be better if this was an array of the tool names, and a loop over those array items with curl + chmod? It would be less duplication and more readable, imho.

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,410 @@
...
+(cd /mnt/data; tar -zcf /tmp/backup/resin-data.tar.gz resin-data)
+
+# Unmount p6
+log "Unmounting filesystems..."
+umount /mnt/data
+umount /var/lib/docker
+umount /resin-data
+
+# Save conf contents to /boot
+# XXX
```
imrehg: What does this `XXX` mean here?

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,410 @@
...
+cp -av /tmp/resin-boot/* /mnt/boot/
+
+# Remove OS image
+rm ${FSARCHIVE}
+
+# Switch root partition
+log "Switching root partition..."
+case $SLUG in
+    beaglebone*)
+	echo 'resin_root_part=3' >/mnt/boot/resinOS_uEnv.txt
```
imrehg: The contents of `resinOS_uEnv.txt` is always this one line? Would it be safer to replace (sed) when that line exists, and only add (with append) if that parameter does not exists? Thinking about any chance that this might lose any data.

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,410 @@
...
+	rm /mnt/boot/uEnv.txt
```
imrehg: Just for info, why is this file removed here?

upgrade-binaries/README.md
```diff
@@ -0,0 +1,12 @@
+# Static binaries for 1.x to 2.x upgrades
+
+Some additional tools are required for upgrading a 1.x resinOS prior to 1.27 to 2.x. These are:
+
+ * e2label
+ * mkfs.ext4
+ * resize2fs
+ * tar
+
+These binaries have been built with buildroot, which makes static builds much easier than Yocto.
```
imrehg: Can you also share the buildroot config that were used to build these binaries?

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,410 @@
...
+
+# Unmount rootB partition
+umount /tmp/rootB
+
+# Extract boot partition, exclude boot_whitelist files
+echo resin-boot/cmdline.txt >/tmp/boot-exclude
+echo resin-boot/config.txt >>/tmp/boot-exclude
+echo resin-boot/splash/resin-logo.png >>/tmp/boot-exclude
+echo resin-boot/uEnv.txt >>/tmp/boot-exclude
+echo resin-boot/EFI/BOOT/grub.cfg >>/tmp/boot-exclude
```
imrehg: Is it for further extension of this script to other device types, or is `resin-boot/EFI/BOOT/grub.cfg` used now on any of the supported devices?

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,410 @@
...
+
+# Export container
+docker export ${CONTAINER} | gzip > ${BACKUPARCHIVE}
+
+# Remove container
+docker rm ${CONTAINER}
+
+# Copy resin-data to backup partition before we wipe data
+cp /mnt/data/resin-data.tar.gz /tmp/backup/
+
```
imrehg: These steps above take quite a bit of time, and very little shown in the log. It would be good, if there was some logged info so one knows where things are. It feels like that those comments could actually be debug log items (similar to existing resinhup scripts), so they inform both when reading and when running the script. The current one sometimes hard to tell what line/step it actually is on.

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,410 @@
...
+	    curl -s -L -o $tools_path/resize2fs $download_uri/resize2fs
+	    chmod 755 $tools_path/resize2fs
+	    ;;
+	*)
+	    log ERROR "Binary type $BINARY_TYPE not supported."
+	    ;;
+    esac
+fi
+
+# Stop docker containers, /var/lib/docker and /resin-data on p6
```
imrehg: It's not exactly what happens in the next step, what the comments here say, isn't it?

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,410 @@
...
+# Unmount backup
+umount /tmp/backup
+
+parted -s $root_dev rm 3
+
+# Resize partition 2 to 327MB
+parted -s $root_dev resizepart 2 373.1
+
+# Create second root partition
+parted -s $root_dev mkpart primary ext4 373 700
```
agherzan: use MiB to make sure we match the images creates by yocto.

upgrade-1.x-to-2.x.sh
agherzan: We should only remove the root partition variable definition if available and maybe mmc dev too (I can't remember exactly what was there before). But there can be things like fdtfile or similar which we don't want to remove as they can be added by the users for various reasons.

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,410 @@
...
+# Copy backup of resin-data
+cp /tmp/backup/resin-data.tar.gz /mnt/data
+
+# Unmount backup
+umount /tmp/backup
+
+parted -s $root_dev rm 3
+
+# Resize partition 2 to 327MB
+parted -s $root_dev resizepart 2 373.1
```
imrehg: Can have some comment/calculation here, how the value 373.1 correspond to that resize that the comment above mentions?

upgrade-1.x-to-2.x.sh
willnewton: It means not implemented. I though this was something that might need doing but I don't really have any further information on it.

upgrade-1.x-to-2.x.sh
willnewton: It's not currently supported. The list is taken from the resinhup config.

upgrade-1.x-to-2.x.sh
willnewton: The file only exists on 2.x AFAIK

agherzan: Correct, it's only in 2.X.

upgrade-1.x-to-2.x.sh
willnewton: Good idea, will do

upgrade-1.x-to-2.x.sh
willnewton: Fixed

upgrade-1.x-to-2.x.sh
willnewton: Fixed

upgrade-1.x-to-2.x.sh
willnewton: I have added a bunch more logging so hopefully it should be a bit clearer now

upgrade-1.x-to-2.x.sh
willnewton: As suggested by Andrei below I've switched this to use sed instead

upgrade-binaries/README.md
willnewton: The key points are:

```
BR2_STATIC_LIBS=y
# BR2_SHARED_LIBS is not set
BR2_TOOLCHAIN=y
BR2_TOOLCHAIN_USES_MUSL=y
BR2_TOOLCHAIN_BUILDROOT=y

BR2_PACKAGE_E2FSPROGS=y
BR2_PACKAGE_E2FSPROGS_E2LABEL=y
BR2_PACKAGE_E2FSPROGS_MKE2FS=y
BR2_PACKAGE_E2FSPROGS_RESIZE2FS=y

BR2_PACKAGE_TAR=y
```

[static-config.txt](https://github.com/resin-os/resinhup/files/1139554/static-config.txt)

upgrade-1.x-to-2.x.sh
willnewton: Fixed

upgrade-1.x-to-2.x.sh
willnewton: I've switched to using sed instead

willnewton:
@agherzan @imrehg the most recent version should take into account your review comments, thanks!

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,422 @@
...
+parted -s $root_dev rm 4
+
+# Create extended partition
+parted -s $root_dev mkpart extended 700MiB 100%
+
+# Create state partition
+parted -s $root_dev mkpart logical ext4 704MiB 724MiB
+
+# Create data partition (using type btrfs corrupts partition table)
+parted -s $root_dev mkpart logical ext4 730MiB 100%
```
agherzan: Should this be starting at 728?

imrehg:
@willnewton running the script I run into an error:

```
[000000027][LOG]Creating new partition table stage 2...
[000000027][LOG]Creating new root filesystems...
resize2fs 1.43.4 (31-Jan-2017)
Filesystem at /dev/mmcblk0p2 is mounted on /; on-line resizing required
old_desc_blocks = 1, new_desc_blocks = 2
resize2fs: Permission denied to resize filesystem
```

willnewton:
@imrehg do you have steps to reproduce?

imrehg:
@willnewton what I did was that I had a 1.26 system that was HUP'd to that version. 

On my retry I couldn't reproduce it, so it's some kind of transient issue?. Basically I did:
```
root@raspberrypi3:~# cd /tmp/
root@raspberrypi3:/tmp# wget https://raw.githubusercontent.com/resin-os/resinhup/8ef1187d2da59c3a66c61979568a7143adf4b075/upgrade-1.x-to-2.x.sh
root@raspberrypi3:/tmp# bash upgrade-1.x-to-2.x.sh
...
```
.... so I guess I keep an eye out?

willnewton:
@imrehg my guess is something had a file open writeable on the rootfs. I try and shut everything down that might do that but its possible I missed something.

upgrade-binaries/README.md
imrehg: Trying to reproduce this, but the latest git from Buildroot does not have for example `BR2_PACKAGE_E2FSPROGS_E2LABEL` config option, and trying `2017.05` (as the config suggest) the config needs adjustments (make menuconfig and then save) to be able to build even, without extra configuration. Checking it more, this is some report of what I've tried.

Regarding the config, I think it would be great of there was a minimal defconfig version, that can be copied into buildroot's `configs`, and made with `make resinhup_static_defconfig && make` or something like that, this would make it reproducible, imho. Not a big deal, but bothering a bit :)

willnewton:
@agherzan i have rebased with changes to make the partition table identical between an upgraded device and a freshly installed 2.x.

upgrade-binaries/README.md
willnewton: The git revision I used was 1a83dda003cc25999589ec392c0b6c81afe7528d

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,437 @@
...
+mount ${root_dev}p3 /tmp/backup
+
+IMAGE=resin/resinos:${TARGET_VERSION}-${DEVICE}
+CONTAINER=export-${TARGET_VERSION}
+BACKUPARCHIVE=/tmp/backup/newos.tar.gz
+FSARCHIVE=/mnt/data/newos.tar.gz
+
+log "Getting new OS image..."
+# Create container for new version
+docker create --name ${CONTAINER} ${IMAGE} echo export
```
petrosagg: this command will return the container id. You can use `CONTAINER=$(docker create ${IMAGE} echo export)` to avoid name conflicts

upgrade-1.x-to-2.x.sh
```diff
@@ -0,0 +1,437 @@
...
+cp /tmp/backup/resin-data.tar.gz /mnt/data
+
+# Unmount backup
+umount /tmp/backup
+
+log "Creating new partition table stage 2..."
+parted -s $root_dev rm 3
+
+# Resize partition 2 to desired size
+# This uses sectors instead of MiB due to a bug in parted. Assumes 512b sectors
```
willnewton: It's the same bug as the one reported here: https://debbugs.gnu.org/cgi/bugreport.cgi/cgi-bin/bugreport.cgi?bug=23511

Upstream parted doesn't seem to be very actively maintained but I submitted a patch anyway.

willnewton:
@petrosagg I've rebased and hopefully taken into account all your suggestions, thanks!

imrehg:
@willnewton I'm guessing the supervisor update after this update has to be done manually when the device rebooted? Or do you think it would it be possible to set up `supervisor.conf` such that on the reboot the device would bring itself up with the new supervisor (instead of manual update)? Thinking similar to https://github.com/resin-os/meta-resin/pull/761 a bit, sans the pull, obviously, something like this, after the rest of the update being successful:

* check the target supervisor version for the release that we are updating to
* set that in the API
* update `supervisor.conf` on the new partition in the correct format (not sure what would that be on 2.x at the moment, too deep in 1.x currently...)

Then on reboot when there's no supervisor there, it will pull the new one automatically (right? this above steps are enough for that to happen?), and the device won't ever run with the new hostOS and old super, but also no need for manually updating the version.

What do you think?
(also cc @agherzan for feedback)

willnewton:
@imrehg from 2.0.8 onwards the update should pull the currently configured supervisor so it would be possible to modify supervisor.conf to pick a new version and update at the same time if we want to.

willnewton:
@agherzan @imrehg are you ok with merging this PR?

agherzan:
Merged. Thanks @willnewton .

PR #77, opened 2017-07-17
updater: allow updating supervisor.conf if the rest of the update succeeds

imrehg:
Update the configuration file when everything is copied to the new partition and it's enabled as the next boot partition (thus everything went well). Change is done through env vars, command line options, or config settings.

### DIFF
--- a/app/modules/updater.py
+++ b/app/modules/updater.py
@@ -202,6 +202,43 @@ class Updater:
             log.debug("Copied " + src + " to " + dst)
         return True
 
+    def updateSupervisorConf(self):
+        """Updating the supervisor.conf on the target rootfs, so after reboot it
+        will have the correct values. The relevant variables are pulled in from
+        the configuration.
+        """
+        log.info("Started to upgrade supervisor conf...")
+        supervisor_image = getConfigurationItem(self.conf, 'Supervisor', 'supervisor_image')
+        supervisor_tag = getConfigurationItem(self.conf, 'Supervisor', 'supervisor_tag')
+        if (not supervisor_image) or (not supervisor_tag):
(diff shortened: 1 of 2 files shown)


app/modules/updater.py
```diff
@@ -202,6 +202,54 @@ def updateBoot(self):
...
+                        for line in supervisorconffile:
+                            # Find and replace supervisor config values
+                            if re.match("^SUPERVISOR_IMAGE=.*$", line):
+                                log.debug("Replacing: SUPERVISOR_IMAGE=" + supervisor_image)
+                                print("SUPERVISOR_IMAGE={}".format(supervisor_image), file=tmpfile)
+                                added_image = True
+                            elif re.match("^SUPERVISOR_TAG=.*$", line):
+                                log.debug("Replacing: SUPERVISOR_TAG=" + supervisor_tag)
+                                print("SUPERVISOR_TAG={}".format(supervisor_tag), file=tmpfile)
+                                added_tag = True
```
willnewton: I think if we replaced the body of the if statements with "pass" it would work the same, i.e. it would get caught below by "if not added_tag"

imrehg: @willnewton I see, you mean not "replace or add", just "add" and then the tags will be always appended to the end of the supervisor.conf, right?

willnewton: Yeah, that should be functionally equivalent but marginally shorter.

imrehg:
@willnewton how's this update?

imrehg:
Probably superseded by https://github.com/resin-os/meta-resin/pull/761 (so the whole update is contained in the bash part of resinhup)

PR #78, opened 2017-07-20
upgrade-2.x.sh: Add script to upgrade 2.x host OS

willnewton:
Add a script that can upgrade a 2.x resinOS version to another
2.x version.

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -1,0 +1,210 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+
+target_version=$1
+preferred_hostos_version=2.0.7
+minimum_target_version=2.0.7
+
+# This will set VERSION and SLUG
+. /etc/os-release
+
+# Log function helper
+function log {
+    # Address log levels
+    case $1 in
+        ERROR)
+            loglevel=ERROR
+            shift
+            ;;
+        WARN)
+            loglevel=WARNING
+            shift
+            ;;
+        *)
+            loglevel=LOG
+            ;;
+    esac
+    endtime=$(date +%s)
+    if [ "z$LOG" == "zyes" ] && [ -n "$LOGFILE" ]; then
+        printf "[%09d%s%s\n" "$((endtime - starttime))" "][$loglevel]" "$1" | tee -a "$LOGFILE"
+    else
+        printf "[%09d%s%s\n" "$((endtime - starttime))" "][$loglevel]" "$1"
+    fi
+    if [ "$loglevel" == "ERROR" ]; then
+        /usr/bin/resin-device-progress --percentage 100 --state "ResinOS: Update failed."
+        exit 1
+    fi
+}
+
+# Test if a version is greater than another
+function version_gt() {
+    test "$(echo "$@" | tr " " "\n" | sort -V | head -n 1)" != "$1"
+}
+
+# Log timer
+starttime=$(date +%s)
+
+# Check board support
+case $SLUG in
+    beaglebone*)
+        binary_type=arm
+        ;;
+    raspberry*)
+        binary_type=arm
+        ;;
+    *)
+        log ERROR "Unsupported board type $SLUG."
+esac
+
+if [ -n "$target_version" ]; then
+    case $target_version in
+        2.*)
+	    if ! version_gt "$target_version" "$minimum_target_version" &&
+		    ! [ "$target_version" == "$minimum_target_version" ]; then
+		log ERROR "Target OS version \"$target_version\" too low, please use \"$minimum_target_version\" or above."
+	    else
+		log "Target OS version \"$target_version\" OK."
+	    fi
+            ;;
+        *)
+            log ERROR "Target OS version \"$target_version\" not supported."
+            ;;
+    esac
+else
+    log ERROR "No target OS version specified."
+fi
+
+# Translate version to one docker will accept as part of an image name
+target_version=$(echo "$target_version" | tr + _)
+
+# Check host OS version
+case $VERSION in
+    2.*)
+        log "Host OS version \"$VERSION\" OK."
+        ;;
+    *)
+        log ERROR "Host OS version \"$VERSION\" not supported."
+        ;;
+esac
+
+# Check if we need to install some more extra tools
+if ! version_gt "$VERSION" "$preferred_hostos_version" &&
+    ! [ "$VERSION" == "$preferred_hostos_version" ]; then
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -0,0 +1,201 @@
+#!/bin/bash
```
michal-mazurek: Bash scripts should have a .bash extension

upgrade-2.x.sh
```diff
@@ -0,0 +1,201 @@
...
+    fi
+}
+
+# Test if a version is greater than another
+function version_gt() {
+    test "$(echo "$@" | tr " " "\n" | sort -V | head -n 1)" != "$1"
+}
+
+# Log timer
+STARTTIME=$(date +%s)
```
michal-mazurek: non-exported functions should be lowercase, to indicate they are non-exported and make the code more readable.

upgrade-2.x.sh
```diff
@@ -0,0 +1,201 @@
...
+
+# Check board support
+case $SLUG in
+    beaglebone*)
+	BINARY_TYPE=arm
```
michal-mazurek: Mixed tabs and spaces for indentation

upgrade-2.x.sh
```diff
@@ -0,0 +1,201 @@
...
+    2.*)
+	log "Host OS version \"$VERSION\" OK."
+	;;
+    *)
+	log ERROR "Host OS version \"$VERSION\" not supported."
+	;;
+esac
+
+# Check if we need to install some more extra tools
+if ! version_gt $VERSION $PREFERRED_HOSTOS_VERSION && ! [ "$VERSION" == $PREFERRED_HOSTOS_VERSION ]; then
```
michal-mazurek: Perhaps break long line

upgrade-2.x.sh
```diff
@@ -0,0 +1,201 @@
...
+
+# Extract boot partition, exclude boot_whitelist files
+log "Extracting new boot partition..."
+echo resin-boot/cmdline.txt >/tmp/boot-exclude
+echo resin-boot/config.txt >>/tmp/boot-exclude
+echo resin-boot/splash/resin-logo.png >>/tmp/boot-exclude
+echo resin-boot/uEnv.txt >>/tmp/boot-exclude
+echo resin-boot/EFI/BOOT/grub.cfg >>/tmp/boot-exclude
+# 2.x adds a default config.json, we should avoid clobbering the existing one
+echo resin-boot/config.json >>/tmp/boot-exclude
```
michal-mazurek: Perhaps more elegant to write like this:
```
( echo
echo
echo) > /tmp/boot-exclude
```

upgrade-2.x.sh
```diff
@@ -0,0 +1,201 @@
...
+	old_root=${root_part#/dev/}
+	new_root=${update_part#/dev/}
+	sed -i -e "s/$old_root/$new_root/" /mnt/boot/cmdline.txt
+	;;
+esac
+
+# Reboot into new OS
+sync
+log "Rebooting into new OS after 5 seconds ..."
+sleep 5
```
michal-mazurek: Why the delay? Perhaps a comment to explain.

willnewton: Mainly to give interested people the chance to Ctrl-C the script if they wish to. ;-) It's probably served it's purpose though and can be removed.

willnewton:
@michal-mazurek I have rebased with changes based on your suggestions (apart from the file extension issue which there doesn't seem to be consensus on at the moment), thanks!

upgrade-2.x.sh
```diff
@@ -0,0 +1,204 @@
...
+        ;;
+    raspberry*)
+        binary_type=arm
+        ;;
+    *)
+        log ERROR "Unsupported board type $SLUG."
+esac
+
+if [ -n "$target_version" ]; then
+    target_version=$(echo "$target_version" | tr + _)
```
agherzan: A small explanation why this translation is needed would help (docker image name restrictions).

upgrade-2.x.sh
```diff
@@ -0,0 +1,204 @@
...
+        *)
+            log ERROR "Target OS version \"$target_version\" not supported."
+            ;;
+    esac
+else
+    log ERROR "No target OS version specified."
+fi
+
+# Check host OS version
+case $VERSION in
```
agherzan: I was little confused about who sets this variable. Maybe a comment that it comes from os-release would help. Or actually when sourcing we could say what variables we are relying on from that file. Just to be clear.

upgrade-2.x.sh
```diff
@@ -0,0 +1,204 @@
...
+        update_part=${root_dev}p2
+        update_part_no=2
+        update_label=resin-rootA
+        ;;
+    *)
+        log ERROR "Unknown root partition \"$root_part\"."
+esac
+
+# Stop docker containers
+log "Stopping all containers..."
```
agherzan: Can we avoid this now? Is supevisor still removing everything? @pcarranzav can we make it ignore resinos images?

upgrade-2.x.sh
```diff
@@ -0,0 +1,204 @@
...
+EOF
+tar -x -X /tmp/root-exclude -C /tmp/updateroot -f $fsarchive
+
+# Extract quirks
+tar -x -C /tmp -f $fsarchive quirks
+cp -a /tmp/quirks/* /tmp/updateroot/
+rm -rf /tmp/quirks
+
+# XXX hack required until 2.0.7 when a quirk is added for this
+(cd /tmp/updateroot/etc; rm mtab; ln -s ../proc/self/mounts mtab)
```
agherzan: Should we only support updates to 2.0.7+? I don't see the benefit in updating to lower versions and keep this hack.

upgrade-2.x.sh
willnewton: If we don't do that then something (I guess the supervisor) cleans up containers which I think can make the export of the container fail below.

agherzan: That is correct. I was wondering if we can make supervisor don't do that.

upgrade-2.x.sh
petrosagg: I've never seen a `.bash` file. Out of all `.sh` files in my system's `/usr`, 40% are bash scripts

michal-mazurek: I posted a rationale for this once. Imagine you have the file open with a text editor. You then cannot execute it by running ./a.sh. So you run sh a.sh. Which then, perhaps only after executing half of the script, crashes with a syntax error.

PR #79, opened 2017-07-26
fixes to the resinhup 1->2 script based on usage

imrehg:
see list of overview in the PR's commit message

### DIFF
--- a/upgrade-1.x-to-2.x.sh
+++ b/upgrade-1.x-to-2.x.sh
@@ -5,9 +5,19 @@ set -o pipefail
 
 . /etc/os-release
 
-MIN_HOSTOS_VERSION=1.16.0
+MIN_HOSTOS_VERSION=1.8.0
 PREFERRED_HOSTOS_VERSION=1.27
+SUPERVISOR_NEW_TAG=v6.1.2
 DOCKER=docker
+
+# Dashboard progress helper
+function progress {
+    percentage=$1
+    message=$2
+    # Progress bar is "nice to have", but might fail on systems
+    # where we move the config.json from it's default location during the update
+    /usr/bin/resin-device-progress --percentage ${percentage} --state "${message}" > /dev/null || true
+}
 
 # Log function helper
 function log {
@@ -32,7 +42,7 @@ function log {
         printf "[%09d%s%s\n" "$(($ENDTIME - $STARTTIME))" "][$loglevel]" "$1"
     fi
     if [ "$loglevel" == "ERROR" ]; then
-        /usr/bin/resin-device-progress --percentage 100 --state "ResinOS: Update failed."
+        progress 100 "ResinOS: Update failed."
         exit 1
     fi
 }
@@ -44,6 +54,8 @@ function version_gt() {
 
 # Log timer
 STARTTIME=$(date +%s)
+
+progress 5 "ResinOS: update prearation..."
 
 # Check board support
 case $SLUG in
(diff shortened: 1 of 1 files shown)


imrehg:
@willnewton cheers! more's coming, that's why I haven't asked for review yet :)

upgrade-1.x-to-2.x.sh
```diff
@@ -76,6 +88,14 @@ else
     esac
 fi
 
+# Find boot path
+boot_path=$(findmnt -n --raw --evaluate --output=target /dev/mmcblk0p1)
+if [ -z "${boot_path}" ]; then
+    log ERROR "Cannot find where is /dev/mmcblk0p1 mounted"
```
agherzan: This is a hard assumption because there will be devices with other block device names. Can we generalise this? In the past we were mounting the boot partition in /boot while now we do it in /mnt/boot. So these are the only mount locations possible for boot partition.

upgrade-1.x-to-2.x.sh
```diff
@@ -160,15 +186,19 @@ umount /var/lib/docker
 umount /resin-data
 
 # Save conf contents to /boot
-# XXX
+if [ -f "/mnt/conf/config.json" ]; then
+    # the boot partition might be mounted ro on some systems, remount
+    mount -o remount,rw "${boot_path}"
```
agherzan: Was that ever the case?

imrehg: When booting 1.8.0 device, I get `ro` mount, tried it a few times, let me check it a bit more.

upgrade-1.x-to-2.x.sh
imrehg: @agherzan sure. Currently there are probably more assumptions as well because of targeting Pi and BB, but one less assumption is better. Will update.

upgrade-1.x-to-2.x.sh
agherzan: I agree there are some more cases here. But I think this one is an easy catch

imrehg: Freshly provisioned `1.8.0`: 
```
|-/boot                          /dev/mmcblk0p1              vfat     ro,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,errors=remount-ro
```
(and no errors in dmesg that would warrant a ro remount happening)

What do you mean by "this one is an easy catch"?

agherzan: Commented to the wrong thread. Fixing..

upgrade-1.x-to-2.x.sh
agherzan: I agree there are some more cases here. But I think this one is an easy catch

upgrade-1.x-to-2.x.sh
imrehg: @agherzan got it, and already pushed changed for your other comment. :)

upgrade-1.x-to-2.x.sh
```diff
@@ -76,6 +90,14 @@ else
     esac
 fi
 
+# Find boot path
+boot_path=$(findmnt -n --raw --evaluate --output=target /dev/${BOOT_PARTITION})
```
willnewton: Can we find by label here? e.g.
findmnt -n --raw --evaluate --output=target LABEL=resin-boot

imrehg:
Added changes based on comments so far.

There's an outstanding issue of the user container being downloaded twice on reboot (once from `registry.`, once from `registry2.`, which should be good to mitigate (have some ideas, and ongoing convo), after that the thing will be complete (for the time being). I will probably add that in another PR, as it might have more feedback required.

So @agherzan @willnewton, this PR should be good to (re-)review, if you'd like (and merge if accepted, that's you, Andrei, right?)

PR #80, opened 2017-07-31
WIP: upgrade-2.x: fixes since merge

imrehg:
Changes in a nutshell:

* fix tool downloads and be more defensive about it.
* do supervisor update in the course of the resinOS update as needed (to have minimum the version provided by the target host OS)
* add progress bars
* only update "prod" devices
* speed up content extraction
* add changes to run the script over ssh & add ssh updater


imrehg:
An outstanding issue, and any feedback is appreciated: because of the bugs in `resin-device-progress`, hopefully fixed in https://github.com/resin-os/meta-resin/pull/791, for resinOS 2.0.8 - 2.2.0 will have to have a patched `resin-device-progress` script provided, otherwise the dashboard won't show any logs. What would be the best way to go about that?
(eg. wait until that patch is merged, and on the affected versions pull the script from github, like the other tools? Or is there a better way?)

imrehg:
@willnewton this was still WIP :/ Sorry that it wasn't completely clear.

willnewton:
@imrehg yeah, sorry about that. A customer was asking about getting access to an experimental 2.x updater script and these changes all looked of a mergeable quality.

imrehg:
@willnewton thanks for the context! Been still testing them a bit, as not sure how stable they are, but they should be pretty close. Will make sure to file anything that comes up so we can have a staging-quality something out.

PR #83, opened 2017-08-09
1.x->2.x updater improvements

imrehg:
Work in progress. So far:

* migrating `resin-wifi-connect` settings alongside `config.json` wifi settings
* bumping target resinOS versions
* adding more logs
* supervisor update done properly
* when migrating the hostOS (in case we are on the second root partition at start) do it safely (ie. read-only remount)
* add command-line options for hostOS versions and supervisor versions
* add command-line options for staging and no-reboot scenarios
* better handling the progress bar for older resinOS versions (where the config is migrated during the update process)
* set minimum target to 2.2.0+rev1 (a new enough hostOS, but not including the breaking change added in 2.3.0 for the bridge IP that some might want to avoid as of now)


upgrade-1.x-to-2.x.sh
```diff
@@ -100,13 +100,13 @@ progress 5 "ResinOS: update prearation..."
 # Check board support
 case $SLUG in
     beaglebone*)
-	TARGET_VERSION=2.0.6_rev2
+	TARGET_VERSION=2.2.0_rev1
```
willnewton: We should now be able to remove the hack around line 473.

imrehg: @willnewton great point! removed :)

imrehg:
@willnewton silly question, i was wondering what was the exact reason to add the NetworkManager config directly into the root-overlay, instead of `resin-boot`? Just wanted to cross check, and not 100% sure :)

willnewton:
@imrehg it is probably just a result of me not understanding that the config is copied across on every boot. I guess it should put it in resin-boot.

imrehg:
@willnewton no worries at all! :D Talked to @josephroberts (being more knowledgeable about networking), and he says it doesn't really matter which place to put it, but `resin-boot` feels "more correct" (paraphrasing). Will change it to that and update the PR :)

imrehg:
Updated PR with latest changes

imrehg:
@agherzan @willnewton this PR is ready to be reviewed too, added all recent changes, and done a few test updates. Much obliged :)

upgrade-1.x-to-2.x.sh
```diff
@@ -56,6 +89,70 @@ function version_gt() {
...
+                        sed -i -e 's/SUPERVISOR_TAG=.*/SUPERVISOR_TAG='"${UPDATER_SUPERVISOR_TAG}"'/' "${supervisor_conf_path}"
+                    else
+                        # Insert supervisor tag
+                        echo "SUPERVISOR_TAG=${UPDATER_SUPERVISOR_TAG}" >> "${supervisor_conf_path}"
+                    fi
+                    # Remove staging registry from the config
+                    sed -i -e 's/SUPERVISOR_IMAGE=registry.resinstaging.io\//SUPERVISOR_IMAGE=/' "${supervisor_conf_path}"
+                    log "Supervisor config update done."
+                else
+                    log ERR "Couldn't extract supervisor vars..."
```
willnewton: I believe this should be ERROR not ERR

imrehg: Updated

imrehg:
Ping @agherzan for a review

upgrade-1.x-to-2.x.sh
```diff
@@ -146,8 +158,15 @@ case $root_part in
...
 	log "Current root partition $root_part is the second root partition. Copying existing OS to first partition..."
     progress 10 "ResinOS: partition switching..."
-	root_dev=${root_part%p3}
+    stop_all
+    sync
+    log "Forcing remount of file systems in read-only mode..."
+    echo u > /proc/sysrq-trigger
+    log "Copying current root partition to the unused partiton..."
+    root_dev=${root_part%p3}
 	dd if=${root_dev}p3 of=${root_dev}p2 bs=4M
```
agherzan: Alignment here seems different.

imrehg:
@agherzan if approved, can you please merge? Thanks!

imrehg:
sorry, need one more timeout on this, found one issue in practice. (wifi migration might fail, and that will just drop the update process), need to put in some more safeguards ASAP for that.

imrehg:
@agherzan @willnewton added fix in the last commit for resin-wifi-connect network migration in the edge case of having a file but not having any config in it.

This should be good to go.... 🍀

imrehg:
Moving this from an external branch to a branch on this repo to work with VersionBot

PR #84, opened 2017-08-31
further improvements to the 2.x->2.x updater

imrehg:
This includes:

* supervisor update automatically to the target resinOS version's shipped supervisor or as defined on the command line with `--supervisor-versions`
* add `resin-device-progress` on devices that shipped with broken version (above 2.6.0 and below 2.3.0, those two versions are good)
* update x86 tools download path
* restart supervisor if the script exits with an error
* other small fixes for shellcheck and scripting

Can squash it in the end, that would be likely better? Any feedback is appreciated. Done more than 100 updates with this setup, seems like getting better and more stable, when I don't do silly things...

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -131,28 +131,31 @@ function upgradeSupervisor() {
         fi
     fi
 
-    CURRENT_SUPERVISOR_VERSION=$(curl -s "${API_ENDPOINT}/v2/device(${DEVICEID})?\$select=supervisor_version&apikey=${APIKEY}" | jq -r '.d[0].supervisor_version')
-    if [ -z "$CURRENT_SUPERVISOR_VERSION" ]; then
-        log ERROR "Could not get current supervisor version from the API, bailing out."
-    else
-        if version_gt "$target_supervisor_version" "$CURRENT_SUPERVISOR_VERSION" ; then
-            log "Supervisor update: will be upgrading from v${CURRENT_SUPERVISOR_VERSION} to ${target_supervisor_version}"
-            UPDATER_SUPERVISOR_TAG="v${target_supervisor_version}"
-            # Get the supervisor id
-            if UPDATER_SUPERVISOR_ID=$(curl -s "${API_ENDPOINT}/v2/supervisor_release?\$select=id,image_name&\$filter=((device_type%20eq%20'$SLUG')%20and%20(supervisor_version%20eq%20'$UPDATER_SUPERVISOR_TAG'))&apikey=${APIKEY}" | jq -e -r '.d[0].id'); then
-                log "Extracted supervisor vars: ID: $UPDATER_SUPERVISOR_ID"
-                log "Setting supervisor version in the API..."
-                curl -s "${API_ENDPOINT}/v2/device($DEVICEID)?apikey=$APIKEY" -X PATCH -H 'Content-Type: application/json;charset=UTF-8' --data-binary "{\"supervisor_release\": \"$UPDATER_SUPERVISOR_ID\"}" > /dev/null 2>&1
-                log "Running supervisor updater..."
-                progress 90 "ResinOS: running supervisor update..."
-                update-resin-supervisor
-                remove_containers
-            else
-                log WARN "Couldn't extract supervisor vars..."
-            fi
+    if CURRENT_SUPERVISOR_VERSION=$(curl -s "${API_ENDPOINT}/v2/device(${DEVICEID})?\$select=supervisor_version&apikey=${APIKEY}" | jq -r '.d[0].supervisor_version'); then
+        if [ -z "$CURRENT_SUPERVISOR_VERSION" ]; then
+            log WARN "Could not get current supervisor version from the API, skipping update..."
         else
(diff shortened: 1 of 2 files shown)


upgrade-2.x.sh
```diff
@@ -290,6 +290,21 @@ if ! version_gt "$VERSION" "$preferred_hostos_version" &&
     esac
 fi
```
willnewton: Should that version be 2.0.6 in the commit message?

upgrade-2.x.sh
```diff
@@ -290,6 +290,21 @@ if ! version_gt "$VERSION" "$preferred_hostos_version" &&
...
 
+# fix resin-device-progress, between version 2.6.0 and 2.3.0
```
willnewton: And here?

upgrade-2.x.sh
```diff
@@ -101,7 +101,7 @@ function remove_containers() {
     docker stop $(docker ps -a -q) > /dev/null 2>&1 || true
     log "Removing all containers..."
     # shellcheck disable=SC2046
-    docker rm -f $(docker ps -a -q) > /dev/null 2>&1 || true
+    docker rm $(docker ps -a -q) > /dev/null 2>&1 || true
```
willnewton: Would it be possible to just drop this patch and the one that added it?

upgrade-2.x.sh
imrehg: Yes, fixed!

imrehg:
@willnewton redone the commits in the PR (removing the one that was reverted, and fixed comment and commit message for another). Thanks for the feedback, let me know if anything else can be improved.

imrehg:
@willnewton @agherzan updated this PR, based on the feedback, plus one more commit to restart the supervisor if the script exits with an error.

Can you please review? :)

imrehg:
Ping @agherzan

imrehg:
Ping @agherzan

upgrade-2.x.sh
agherzan: @willnewton I don't quite understand your comment here.

upgrade-2.x.sh
```diff
@@ -124,7 +133,7 @@ function upgradeToReleaseSupervisor() {
                     log "Running supervisor updater..."
                     progress 90 "ResinOS: running supervisor update..."
                     update-resin-supervisor
-                    stop_services
+                    remove_containers
```
agherzan: NIT: I would find it a little clearer if I had here both stop and remove functions invoked. I understand the dependency which made you code it this way though.

upgrade-2.x.sh
imrehg: @agherzan there were two commits before, one added the "-f" function, and another removed it (thus the line was unchanged in the end). Thus the comment was to drop those two commits altogether instead of having a patch and a revert in a WIP PR. So you are replying to an outdated state.

upgrade-2.x.sh
```diff
@@ -102,44 +108,50 @@ function upgradeToReleaseSupervisor() {
...
+                log "Extracted supervisor vars: ID: $UPDATER_SUPERVISOR_ID"
+                log "Setting supervisor version in the API..."
+                curl -s "${API_ENDPOINT}/v2/device($DEVICEID)?apikey=$APIKEY" -X PATCH -H 'Content-Type: application/json;charset=UTF-8' --data-binary "{\"supervisor_release\": \"$UPDATER_SUPERVISOR_ID\"}" > /dev/null 2>&1
+                log "Running supervisor updater..."
+                progress 90 "ResinOS: running supervisor update..."
+                update-resin-supervisor
+                remove_containers
             else
-                log "Supervisor update: no update needed."
+                log WARN "Couldn't extract supervisor vars..."
```
agherzan: Don't we want to error out in this case?

upgrade-2.x.sh
```diff
@@ -102,44 +108,50 @@ function upgradeToReleaseSupervisor() {
...
-                    progress 90 "ResinOS: running supervisor update..."
-                    update-resin-supervisor
-                    remove_containers
-                else
-                    log WARN "Couldn't extract supervisor vars..."
-                fi
+        if version_gt "$target_supervisor_version" "$CURRENT_SUPERVISOR_VERSION" ; then
+            log "Supervisor update: will be upgrading from v${CURRENT_SUPERVISOR_VERSION} to ${target_supervisor_version}"
+            UPDATER_SUPERVISOR_TAG="v${target_supervisor_version}"
+            # Get the supervisor id
```
agherzan: It would be really helpful to add a comment explaining what this ID does.

upgrade-2.x.sh
imrehg: Updated!

imrehg:
@agherzan @willnewton added changes based on feedback (see last two commits)

If this is approved, please merge :)

imrehg:
Ping @agherzan @willnewton for review (was cleared by latest commit cleanup)

imrehg:
Moving from external branch to a branch on this repo to work with VersionBot

PR #87, opened 2017-09-18
upgrade-2.x: changes in preparation to self-service hup

imrehg:
The set of changes in this PR:

* The logging function in this file was already able to handle logging to a file (eg. for when the script is invoked through the proxy). Add the relevant setup section and command line flags to actually make use of it.
* being able to use staging resinOS docker hub
* switching order of supervisor update with rootfs switch for bringing upgrade changes closer to each other (and be reversible if the supervisor update fails)
* adding logic to handle the root of Beaglebone and NUC devices properly
* add better logging that saves all logs into the file (including standard output and errors)
* checking if image:tag exists in registry (docker hub) and bail out if not
* remove vulnerable `resin-sample` from the device if present on update #93 
* fix logic of getting APIKEY for some versions where both apikey and deviceapikey is present (probably >=2.0.0 || <2.0.3)
* sanity check for device types

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -224,7 +224,7 @@ function remove_sample_wifi {
     # connection settings, as they are well known and thus insecure
     local filename=$1
     if [ -f "${filename}" ] && grep -Fxq "ssid=My_Wifi_Ssid" "${filename}" && grep -Fxq "psk=super_secret_wifi_password" "${filename}" ; then
-        if nmcli | grep "resin-sample"; then
+        if nmcli c  show --active | grep "resin-sample" ; then
             # If a connection with that name is in use, do not actually remove the settings
             log WARN "resin-sample configuration found at ${filename} but it might be connected, not removing..."
         else
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -420,13 +420,9 @@ if [ -f /mnt/boot/config.json ]; then
 else
     log ERROR "Don't know where config.json is."
 fi
-APIKEY=$(jq -r '.deviceApiKey' $CONFIGJSON)
-if [ "$APIKEY" == 'null' ]; then
-    log WARN "Using apiKey as device does not have deviceApiKey yet..."
-    APIKEY=$(jq -r '.apiKey' $CONFIGJSON)
-fi
-DEVICEID=$(jq -r '.deviceId' $CONFIGJSON)
-API_ENDPOINT=$(jq -r '.apiEndpoint' $CONFIGJSON)
+# If the user api key exists we use it instead of the deviceApiKey as it means we haven't done the key exchange yet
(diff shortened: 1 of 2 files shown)


upgrade-2.x.sh
```diff
@@ -45,6 +47,9 @@ Options:
         If not defined, then the update will try to run for the HOSTOS_VERSION's
         original supervisor release.
 
+    -n, --nolog
```
agherzan: I would rename the flag to be clear that it only deactivates log to file.

imrehg: @agherzan very much agree, this form is to keep it the same as the 1.x->1.x flags, which surprised me with this slightly off meaning (see https://github.com/resin-os/meta-resin/blob/1.X/meta-resin-common/recipes-support/resinhup/resinhup/run-resinhup.sh#L68-L69 )

So I wonder, is it better to take this chance to break uniformity between scripts, or not?

agherzan: Uh I see. OK. Let's leave it as it is for now.

imrehg:
@agherzan can you plz merge?

imrehg:
@agherzan plz merge, if you have a chance?

imrehg:
ping @agherzan can you please merge this (so I can PR other changes that build on top of these changes?) Thanks!

imrehg:
Dismissing reviews, and just going to add my changes on top of this, so only one review/merge will be needed.

imrehg:
Ready for review / approval / merge.

imrehg:
@CameronDiver when running `nmcli` without arguments, only the connected profiles are listed, e.g. `wlan0: connected to resin-sample`, and no other states show up. Thus I don't think we can catch any other states. That line is trying to catch when that profile is in use, for better or worse, because we shouldn't remove a profile that might leave the device bricked/unable to connect.

Having said that, a more explicit way of doing this would be `nmcli c show --active | grep resin-sample`, which definitely just lists the active connection. Will try it out, and should probably switch over to that.

upgrade-2.x.sh
```diff
@@ -224,7 +224,7 @@ function remove_sample_wifi {
     # connection settings, as they are well known and thus insecure
     local filename=$1
     if [ -f "${filename}" ] && grep -Fxq "ssid=My_Wifi_Ssid" "${filename}" && grep -Fxq "psk=super_secret_wifi_password" "${filename}" ; then
-        if nmcli | grep "resin-sample"; then
+        if nmcli c  show --active | grep "resin-sample" ; then
```
imrehg: @CameronDiver revision done ^^

imrehg:
@agherzan can this be reviewed and merged? this is the form that got into self-service v2, and have further improvements that I'd like to add, but the PR is getting waaaay too big I had to pile on it due to not having recent merges.

agherzan:
@imreh done.

PR #88, opened 2017-10-04
ugrade-1.x-to-2.x: changes to go into the self-service HUP

imrehg:
* Exit properly when switching root partition
* Source profile to get proper path information
* Enable loging (to /tmp/ as for 1.x->2.x the only filesystem  that does not
  change in the process)
* Allow pulling from the staging resinOS Docker HUB repository
* Add script to call the updater through ssh (optional, for manual updates)


imrehg:
Bummer, pushed to my own fork, which does not work with versionist, repushing....

PR #89, opened 2017-10-04
ugrade-1.x-to-2.x: changes to go into the self-service HUP

imrehg:
* Exit properly when switching root partition
* Source profile to get proper path information
* Enable loging (to /tmp/ as for 1.x->2.x the only filesystem  that does not
  change in the process)
* Allow pulling from the staging resinOS Docker HUB repository
* Add script to call the updater through ssh (optional, for manual updates)
* Increase minimum hostos starting version for BB, so that we avoid the eMMC bug
* Bail out better if the resinos image download fails (instead of silently dying)
* Better logging on the device
* Checking if image:tag exists in registry (docker hub) and bail out if not
* Fix start-resin-supervisor for versions that have broken logic
* Log error properly when resin-data backup fails (e.g. if backup larger than available space to back up)
* For releases that come with potentialy broken supervisor (6.2.0-6.3.4), upgrade to the next available supervisor (6.3.5)

### DIFF
--- a/upgrade-1.x-to-2.x.sh
+++ b/upgrade-1.x-to-2.x.sh
@@ -592,6 +592,9 @@ log "Getting new OS image..."
 progress 50 "ResinOS: downloading OS update..."
 # Create container for new version
 CONTAINER=$(docker create ${IMAGE} echo export)
+if [ -z "$CONTAINER" ]; then
+    log ERROR "Could not download target update image..."
+fi
 
 progress 60 "ResinOS: processig update package..."
 # Export container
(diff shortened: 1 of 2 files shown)


imrehg:
@agherzan can this be reviewed and merged? this is the form that got into self-service v2, and have further improvements that I'd like to add, but the PR is getting waaaay too big I had to pile on it due to not having recent merges.

agherzan:
Let's see...

upgrade-1.x-to-2.x.sh
```diff
@@ -203,6 +203,48 @@ function stop_all() {
     systemctl stop docker
 }
 
+function image_exsits() {
```
agherzan: There is a typo here.

imrehg: Thanks, fixed!

PR #106, opened 2017-11-01
upgrade-1.x-to-2.x: set to be deleted partitions to read only and flush buffer

imrehg:
In some rare, and hard-to-reproduce cases, when parted removes a partition,the partition table information fails to update, because maybe the partition is somehow still in use even if it was unmounted successfully.

Trying to prevent that case by setting the soon-to-be-removed partitions to read-only and flushing the device buffers, before removing the partition.

### DIFF
--- a/upgrade-1.x-to-2.x.sh
+++ b/upgrade-1.x-to-2.x.sh
@@ -526,6 +526,11 @@ fi
 # Unmount p1
 umount "${boot_path}"
 
+# Set soon-to-be-removed block devices to read-only, and flush the buffers
+blockdev --setro ${root_dev}p6
+blockdev --setro ${root_dev}p5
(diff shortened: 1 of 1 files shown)


imrehg:
Proposed. So far to me it seems this change can not hurt; in the worst case scenario this does not make any difference and will still hit the issue. Ideally we'll plug some race condition that results in the error we've seen (but can't yet reproduce on demand)

> Error: Partition(s) 6 on /dev/mmcblk0 have been written, but we have been unable to inform the kernel of the change, probably because it/they are in use. As a result, the old partition(s) will remain in use. You should reboot now before making further changes.

imrehg:
Does not fix the issue, looking for other solution.

Issue #107, opened 2017-11-03
update-resin-supervisor polyfill for some versions


imrehg:
update-resin-supervisor does not properly work with device API keys for some versions (e.g 2.2.0->2.3.0), and add polyfill for that (thought it was already added, but looks like maybe not). The lack of this polyfill would stopresinHUP updating some versions.

PR #108, opened 2017-11-03
upgrade-2.x: update-resin-supervisor polyfill for 2.0.8-2.7.0 versions

imrehg:
These versions had update-resin-supervisor that couldn't properly query the API for an update.

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -467,6 +467,22 @@ else
     log "No resin-device-progress fix is required..."
 fi
 
+# Fix for issue: https://github.com/resin-os/meta-resin/pull/864
+# Also includes change from: https://github.com/resin-os/meta-resin/pull/882
+if version_gt "$VERSION_ID" "2.0.8" &&
+    version_gt "2.7.0" "$VERSION_ID"; then
+        log "Fixing supervisor updater..."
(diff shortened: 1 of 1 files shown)


imrehg:
Tested by updating 2.2.0+rev1 -> 2.7.5+rev1 (with and without the patch)

imrehg:
Fixes #107

imrehg:
@agherzan can I have a review and merge if good?

imrehg:
Ping @agherzan ?

upgrade-2.x.sh
```diff
@@ -467,6 +467,20 @@ else
...
 
+if version_gt "$VERSION_ID" "2.0.8" &&
+    version_gt "2.7.0" "$VERSION_ID"; then
+        log "Fixing supervisor updater..."
+        if curl --fail --silent -o "/tmp/update-resin-supervisor" https://raw.githubusercontent.com/resin-os/meta-resin/40d5a174da6b52d530c978e0cae22aa61f65d203/meta-resin-common/recipes-containers/docker-disk/docker-resin-supervisor-disk/update-resin-supervisor ; then
+            chmod 755 "/tmp/update-resin-supervisor"
+            PATH="/tmp:$PATH"
+            log "Added temporary supervisor updater replaced with fixed version..."
+        else
+            log WARN "Could not download temporary supervisor updater..."
```
floion: Do you not want it to error out in this case? Or at least enhance the warning saying this really needs to be fixed on the device afterwards?

imrehg: @floion yes, probably should, will update it like that. The stronger warning wouldn't make a difference, because logs from self-service updates will just stay on the device now, and won't get read by us, unless we need to see how things went.

imrehg:
Updated from Florin's comment

PR #112, opened 2017-11-16
upgrade-2x: enable hostapp-update based updates through resinhup

imrehg:
* whenever available, use hostapp-update
* do pre-update cleanup
* do pre-update fixes/polyfills
* work on both regular 2.x->2.x updates and the newer, hostapp-update ones

Connects-to: https://github.com/resin-io/hq/pull/1087
Change-type: minor

Tested on rpi3 for many different versions, bbb, up-board, intel-nuc.


upgrade-2.x.sh
```diff
@@ -249,6 +249,70 @@ function device_type_match {
     echo "${match}"
 }
 
+# Pre update cleanup: remove some not-required files from the boot partition to clear some space
+function pre_update_pi_bootfiles_removal {
+    local boot_files_for_removal=('start_db.elf' 'fix_db.dat')
```
floion: This should be "fixup_db.dat"

upgrade-patches/0-bootfiles
```diff
@@ -0,0 +1,127 @@
...
+bootfiles_list="/usr/share/hostapp-update/bootfiles.txt"
+boot_fingerprint="resinos.fingerprint"
+bootfiles_blacklist="\
+	/config.json \
+	/config.txt \
+	/splash/resin-logo.png \
+	"
+bootfiles_ignore="\
+	/start_db.elf \
+	/fix_db.dat \
```
floion: This should be "fixup_db.dat"

imrehg: Addressed, thanks!

upgrade-2.x.sh
imrehg: Addressed, thanks!

bbinet:
Could you explain what is hostapp-update ?
I can't find any information about it.
Thanks.

imrehg:
@floion @agherzan can I have a review here?

imrehg:
@bbinet sure thing, will do in a bit, it's an upcoming resinOS feature that we are very excited about, and makes updates even better.

bbinet:
@imrehg Great, I'm looking forward to hear about it!

imrehg:
Implemented the migration as discussed / spec'd out, should be ready for review (testing in the meantime)

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
+    local update_package=$1
+    local sysroot="/mnt/sysroot/inactive"
+    local temp_sysroot
+
+    find_partitions
+    stop_services
+
+    # Mount spare root partition
+    umount "${update_part}" || true
+    mkfs.ext4 -F -L "${update_label}" "${update_part}"
```
petrosagg: Use the same parameters we use in resinOS when creating partitions: https://github.com/resin-os/meta-resin/blob/b32f8be8cd90d39732c707f009213e616a6396fa/meta-resin-common/recipes-containers/mkfs-hostapp-native/files/create#L16

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
+    docker run \
+      --rm \
+      --name resinhup \
+      --privileged \
+      -v "${temp_sysroot}":/mnt/sysroot/inactive \
+      -v /mnt/boot:/mnt/boot \
+      -v /tmp/resinos-image.docker:/resinos-image.docker \
+      -v /mnt/data/resinhup/tmp:/mnt/data/resinhup/tmp \
+      "${update_package}" \
+      /bin/bash -c 'storage_driver=$(cat /boot/storage-driver) ; DOCKER_TMPDIR=/mnt/data/resinhup/tmp/ dockerd --log-driver=journald -s=$storage_driver --data-root='"${sysroot}"'/docker -H unix:///var/run/docker-host.sock --pidfile=/var/run/docker-host.pid --exec-root=/var/run/docker-host --bip 10.114.201.1/24 --fixed-cidr=10.114.201.128/25 --iptables=false & sleep 10; echo "Starting hostapp-update"; hostapp-update -f /resinos-image.docker' \
```
petrosagg: Use `until docker ps; do sleep 0.2; done` here too

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
+    || log ERROR "Update based on hostapp-update has failed..."
+}
+
+function find_partitions {
+    # Find which partition is / and which we should write the update to
+    root_part=$(findmnt -n --raw --evaluate --output=source /)
+    log "Found root at ${root_part}..."
+    case $root_part in
+        *2)
+            # on 2.x the following device types have these kinds of results for $root_part
```
petrosagg: @agherzan this is a bug. Why are not all device use label based partitions?

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
+      /bin/bash -c 'storage_driver=$(cat /boot/storage-driver) ; DOCKER_TMPDIR=/mnt/data/resinhup/tmp/ dockerd --log-driver=journald -s=$storage_driver --data-root='"${sysroot}"'/docker -H unix:///var/run/docker-host.sock --pidfile=/var/run/docker-host.pid --exec-root=/var/run/docker-host --bip 10.114.201.1/24 --fixed-cidr=10.114.201.128/25 --iptables=false & sleep 10; echo "Starting hostapp-update"; hostapp-update -f /resinos-image.docker' \
+    || log ERROR "Update based on hostapp-update has failed..."
+}
+
+function find_partitions {
+    # Find which partition is / and which we should write the update to
+    root_part=$(findmnt -n --raw --evaluate --output=source /)
+    log "Found root at ${root_part}..."
+    case $root_part in
+        *2)
```
petrosagg: This index based approach will not work on boards like the edison or the jetson tx2 that do not have their root partition on offsets 2 or 3

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
+            update_part_no=3
+            update_label=resin-rootB
+            ;;
+        *3)
+            root_dev=${root_part%3}
+            update_part=${root_dev}2
+            update_part_no=2
+            update_label=resin-rootA
+            ;;
+        *rootA)
```
petrosagg: This can only be `resin-rootA`, right? If yes, no need for the wildcard

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
+            old_label=resin-rootA
+            update_label=resin-rootB
+            root_part_dev=$(blkid | grep "${old_label}" | awk '{print $1}' | sed 's/://')
```
petrosagg: Can we do this with `findmnt` or with `$(readlink -f /dev/disk/by-label/${old_label})`?

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
+    fi
+}
+
+function finish_up() {
+    sync
+    if [ "${NOREBOOT}" == "no" ]; then
+        # Reboot into new OS
+        log "Rebooting into new OS in 5 seconds..."
+        progress 100 "ResinOS: update successful, rebooting..."
+        nohup bash -c " /bin/sleep 5 ; /sbin/reboot " > /dev/null 2>&1 &
```
petrosagg: Why not just `sleep 5; reboot`?

upgrade-2.x.sh
```diff
@@ -483,56 +656,40 @@ else
...
-        update_part=${root_part_dev%2}3
-        ;;
-    *)
-        log ERROR "Unknown root partition \"$root_part\"."
-esac
-if [ ! -b "$update_part" ]; then
-    log ERROR "Update partition detected as ${update_part} but it's not a block device."
+### hostapp-update based updater
+
+if [ -x "$(command -v hostapp-update)" ]; then
```
petrosagg: Why are we not using the OS version for this? `minimum_hostapp_target_version` is set at the top of the script

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
+        storage_driver=$(cat /boot/storage-driver)
+        log "Starting docker-host with ${storage_driver} storage driver"
+        dockerd --log-driver=journald -s="${storage_driver}" --data-root="${sysroot}/docker" -H unix:///var/run/docker-host.sock --pidfile=/var/run/docker-host.pid --exec-root=/var/run/docker-host --bip 10.114.101.1/24 --fixed-cidr=10.114.101.128/25 --iptables=false &
+        sleep 10
+    else
+        if [ -f "$sysroot/resinos.fingerprint" ]; then
+            # Happens on a device, which has HUP'd from a non-hostapp resinOS first
+            # then updated again but this time with hostapp-update. The previous "active"
+            # partition now inactive, and still has leftover data
+            log "Have docker-host running, with dirty inactive partition"
```
petrosagg: How can this happen? In a hostapp to hostapp update there should be no garbage left behind that need to be cleaned up after the new OS has booted

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
+    if ! [ -S "/var/run/docker-host.sock" ]; then
+        ## Happens on devices booting after a regular HUP update onto a hostapps enabled resinOS
+        log "Do not have docker-host running"
+        log "Clean inactive partition"
+        rm -rf "${sysroot:?}/"*
+        log "Getting storage driver info"
+        storage_driver=$(cat /boot/storage-driver)
+        log "Starting docker-host with ${storage_driver} storage driver"
+        dockerd --log-driver=journald -s="${storage_driver}" --data-root="${sysroot}/docker" -H unix:///var/run/docker-host.sock --pidfile=/var/run/docker-host.pid --exec-root=/var/run/docker-host --bip 10.114.101.1/24 --fixed-cidr=10.114.101.128/25 --iptables=false &
+        sleep 10
```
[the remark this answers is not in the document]
imrehg: done

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
+        if [ -f "$sysroot/resinos.fingerprint" ]; then
+            # Happens on a device, which has HUP'd from a non-hostapp resinOS first
+            # then updated again but this time with hostapp-update. The previous "active"
+            # partition now inactive, and still has leftover data
+            log "Have docker-host running, with dirty inactive partition"
+            systemctl stop docker-host
+            log "Clean inactive partition"
+            rm -rf "${sysroot:?}/"*
+            systemctl start docker-host
+            sleep 10
```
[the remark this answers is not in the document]
imrehg: done

upgrade-2.x.sh
imrehg: Done

upgrade-2.x.sh
imrehg: done

upgrade-2.x.sh
imrehg: updated

upgrade-2.x.sh
imrehg: @petrosagg you mean instead of absolute path, or instead of using "nohup"? I take the former, but let me know if I miss something. The nohup I think is there for properly `exit 0` from the script (but it's a holdover from the original version I received, so happy to change it it could be improved further.

upgrade-2.x.sh
imrehg: I fell this is more robust to check whether the current system has hostapp-update support or not.. The `minimum_hostapp_target_version` is checking the update target, while this check is checking the current system. Can update to use the version check too, then.

upgrade-2.x.sh
imrehg: Updated.

upgrade-2.x.sh
imrehg: This function is only used for pre-hostapp-update devices, so for the Jetson TX2 it should not cause issue. For the Edison this will be an issue, when adding support for HUP in the near future. In that case will add Edison-specific stuff in this logic.

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
+        rm -f "/mnt/boot/$f"
+    done
+    sync
+}
+
+function pre_update_fix_bootfiles_hook {
+    log "Applying bootfiles hostapp-hook fix"
+    local bootfiles_temp
+    bootfiles_temp=$(mktemp)
+    curl -f -s -L -o "$bootfiles_temp" https://raw.githubusercontent.com/resin-os/resinhup/master/upgrade-patches/0-bootfiles || log ERROR "Couldn't download fixed '0-bootfiles', aborting."
```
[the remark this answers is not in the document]
imrehg: Broke that file out into its own PR, #115, so when that's merged, can refer to it by commit.

upgrade-2.x.sh
imrehg: @petrosagg it's not the hostapp-to-hostapp updatre case, but as the header says, that non-hostapp-to-hostapp. There the previous system is still there, as before the reboot it was running, and currently don't have a service in resinOS that clears up `/mnt/sysroot/inactive` on boot (before starting the `docker-host` service).

upgrade-2.x.sh
agherzan: I think the problem here is to detect which root we are running on. They all use labels but the problem is detecting if we are running from rootA or rootB.

upgrade-2.x.sh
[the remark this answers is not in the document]
agherzan: It was merged :)

upgrade-2.x.sh
```diff
@@ -249,6 +250,176 @@ function device_type_match {
...
 }
 
+# Pre update cleanup: remove some not-required files from the boot partition to clear some space
+function pre_update_pi_bootfiles_removal {
+    local boot_files_for_removal=('start_db.elf' 'fixup_db.dat')
+    for f in "${boot_files_for_removal[@]}"; do
+        echo "Removing $f from boot partition"
+        rm -f "/mnt/boot/$f"
+    done
+    sync
```
agherzan: You can sync just the boot partition. `sync /mnt/boot`

upgrade-2.x.sh
```diff
@@ -249,6 +250,212 @@ function device_type_match {
...
+    if ! [ -S "/var/run/docker-host.sock" ]; then
+        ## Happens on devices booting after a regular HUP update onto a hostapps enabled resinOS
+        log "Do not have docker-host running"
+        log "Clean inactive partition"
+        rm -rf "${sysroot:?}/"*
+        log "Getting storage driver info"
+        storage_driver=$(cat /boot/storage-driver)
+        log "Starting docker-host with ${storage_driver} storage driver"
+        dockerd --log-driver=journald -s="${storage_driver}" --data-root="${sysroot}/docker" -H unix:///var/run/docker-host.sock --pidfile=/var/run/docker-host.pid --exec-root=/var/run/docker-host --bip 10.114.101.1/24 --fixed-cidr=10.114.101.128/25 --iptables=false &
+        until DOCKER_HOST="unix:///var/run/docker-host.sock" docker ps &> /dev/null; do sleep 0.2; done
```
agherzan: We should try this with a timeout otherwise if host docker can't start we will end up blocked.

upgrade-2.x.sh
petrosagg: Right. I thought I had deleted this comment before submitting my review. Thanks for clarifying anyway

upgrade-2.x.sh
petrosagg: got it

upgrade-2.x.sh
imrehg: Done

upgrade-2.x.sh
imrehg: @agherzan check if the changes are good for you.

I've tested all 3 of such calls, both when everything worked, and when I changed the script to not start up docker and thus see that it correctly failed/timed-out.
Any feedback's appreciated.

PR #118, opened 2017-12-07
upgrade-1.x-to-2.x: check if btrfs partition was successfully umounted

imrehg:
In some circumstances, it seems like there are lingering processes after the data partition, which is btrfs on resinOS 1.x, is umounted. When that's the case, continuing to the next step is not desirable, as the following steps are trying to use the partition that is still in use, and will fail, resulting in a broken-modified device file system.

In these cases, manual resinHUP is needed, most likely, to update the device. So here just bail out before anything is changed in the first test. Then the device can be manually rebooted, and will come online as was before.

The second test is less likely to fail (as it's a newly created btrfs partition), but it might, so better test there too.

### DIFF
--- a/upgrade-1.x-to-2.x.sh
+++ b/upgrade-1.x-to-2.x.sh
@@ -318,6 +318,19 @@ function fix_supervisor_bootstrap {
     fi
 }
 
+function check_btrfs_umount() {
+    # Check whether /data has been correctly umounted, which is the only btrfs
+    # partition;  if not, then bail out before anything's destroyed
+    local timeout_seconds=$((SECONDS+30));
+    while pidof btrfs-worker > /dev/null; do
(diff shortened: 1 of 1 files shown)


upgrade-1.x-to-2.x.sh
```diff
@@ -563,6 +563,16 @@ umount /mnt/data
 umount /var/lib/docker
 umount /resin-data
 
+# Check whether /data has been correctly umounted, and stop if not, before anything's destroyed
```
agherzan: Should this be a function? Looks identical as the next check.

imrehg: You are right, definitely!

imrehg:
@agherzan updated and retested, ready for review

PR #122, opened 2017-12-12
upgrade-1x-to-2x/upgrade-2x: update progress messages based on feedback

imrehg:


upgrade-1.x-to-2.x.sh
```diff
@@ -492,7 +492,7 @@ case $root_part in
                 ;;
         esac
         log "Rebooting..."
-        progress 15 "ResinOS: rebooting to continue..."
+        progress 15 "Rebooting from switched partition"
```
alisondavis17: maybe "second" or "new" would be clearer than "switched?" it's not clear which partition is meant by "switched"

upgrade-1.x-to-2.x.sh
```diff
@@ -740,14 +740,14 @@ BACKUPARCHIVE=/tmp/backup/newos.tar.gz
...
-progress 50 "ResinOS: downloading OS update..."
+progress 50 "Downloading OS update"
 # Create container for new version
 CONTAINER=$(docker create ${IMAGE} echo export)
 if [ -z "$CONTAINER" ]; then
     log ERROR "Could not download target update image..."
 fi
 
-progress 60 "ResinOS: processing update package..."
+progress 60 "Processing OS update"
```
alisondavis17: maybe "processing OS update package"?

upgrade-1.x-to-2.x.sh
imrehg: Neither (the pr or the alternatives) feel really good... will keep thinking.

upgrade-1.x-to-2.x.sh
imrehg: Yeah, will update it. That will make it pretty much as long as the original was, wonder if there's any more concise way to express something similar.

upgrade-1.x-to-2.x.sh
```diff
@@ -888,10 +888,10 @@ sync
...
     # Reboot into new OS
     log "Rebooting into new OS in 5 seconds..."
-    progress 100 "ResinOS: update successful, rebooting..."
+    progress 100 "Update successful, rebooting"
     nohup bash -c " /bin/sleep 5 ; /sbin/reboot " > /dev/null 2>&1 &
 else
     log "Finished update, not rebooting as requested."
     log "NOTE: Supervisor and stopped services kept stopped!"
-    progress 100 "ResinOS: update successful."
+    progress 100 "Update successful."
```
agherzan: I would remove the full stop from this too as you seem to have removed them from all the other messages.

imrehg: Yup, that's good.

imrehg:
@alisondavis17 @agherzan updated based on suggestions, thanks a bunch!

PR #137, opened 2018-01-17
upgrade-2.x: remove supervisor database after upgrade for some versions

imrehg:
Some supervisor versions are not great at database migration, and after HUP might not start up properly. Remove the database for the affected versions' updates as a workaround. Fixed such issues in supervisor v6.5.7 and v6.5.9.

Change-type: patch

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -198,6 +198,10 @@ function upgrade_supervisor() {
                     update-resin-supervisor
                     stop_services
                     remove_containers
+                    if version_gt "6.5.9" "${target_supervisor_version}" ; then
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -198,6 +198,10 @@ function upgrade_supervisor() {
                     update-resin-supervisor
                     stop_services
                     remove_containers
+                    if version_gt "6.5.9" "${target_supervisor_version}" ; then
```
agherzan: Please try to reference issues, PR or other resources that would show the initial issue and how it was fixed (if at all).

PR #143, opened 2018-02-02
upgrade-2.x: fix balena migration for non-hostapp-mode hostapp enabled resinOS

### LINKED ISSUE #142: Properly handle balena migration of non-hostapp-running 2.7.5 variants
There are some devices that are running hostapp-capable 2.7.5 resinOS, but not in proper hostapp-mode. Those need some migration on subsequent updates. The current script doesn't handle yet if the next update is a balena migration.

Symptom is resinHUP log with something like:

```
[ERROR]docker-host did not come up before check timed out...
```

Front logo [Front conversations](https://app.frontapp.com/open/top_7hq1)

imrehg:
Previous resinHUP changes have updated devices to versions that run in hostapp mode, but due to the incomplete logic of the update, the devices ended up running the new system in the "regular", old mode outside of hostapp.

Fixes #142

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -6,6 +6,7 @@ IGNORE_SANITY_CHECKS=no
 RESINOS_REGISTRY="registry.hub.docker.com"
 RESINOS_REPO="resin/resinos"
 SCRIPTNAME=upgrade-2.x.sh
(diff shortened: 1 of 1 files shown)


imrehg:
Added more info in commit message based on feedback.

PR #148, opened 2018-03-08
upgrade-2.x.sh: fix issues with `read` on newer resinOS versions

imrehg:
The read command only passes the first variable starting 2.9.x, thus we cannot use it anymore to do multiple shell variable set.

Replace that call, and patch `update-resin-supervisor` which is also affected.

Change-type: patch

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -765,8 +765,9 @@ else
     log ERROR "Don't know where device-type.json is."
 fi
 # If the user api key exists we use it instead of the deviceApiKey as it means we haven't done the key exchange yet
-# shellcheck disable=SC2046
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -895,6 +896,15 @@ else
     log "No supervisor updater fix is required..."
 fi
 
+# Fix issue with `read` on 2.10.x/2.11.0 resinOS versions
+if version_gt "$VERSION_ID" "2.9.7"; then
```
agherzan: We should avoid applying for versions which will have it fixed.

imrehg: @agherzan changed, and that was the plan, just didn't know if it will be fixed.

PR #170, opened 2018-06-04
updater for a resinOS variant sometimes (rarely) found in production

imrehg:
* Fixing the state partition location and label.
* Use correct slug and sidestep sanity check as we'd know the results

Change-type: patch

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -622,6 +622,26 @@ function find_partitions {
 }
 
 #######################################
+# In some cases, the state partition does not have the required `resin-state` label
+# and upon reboot the device would be inoperable. Apply that label as needed.
+# Globals:
+#   root_dev
+#   state_part
+# Returns:
+#   None
+#######################################
+function fix_state_label() {
+    if [ "${state_part}" != "/mnt/state" ]; then
+        log "Need to apply proper state partition label."
+        local state_dev
+        state_dev=$(findmnt -n --raw --evaluate --output=source "${state_part}")
+        if [ -n "${state_dev}" ]; then
+            e2label "${state_dev}" "resin-state"
+        fi
+    fi
+}
+
+#######################################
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -623,6 +623,22 @@ function find_partitions {
...
+# Globals:
+#   root_dev
+#   state_part
+# Returns:
+#   None
+#######################################
+function fix_state_label() {
+    if [ "${state_part}" != "/mnt/state" ]; then
+        log "Need to apply proper state partition label."
+        e2label ${root_dev}5 "resin-state"
```
agherzan: Do we have boards with different partition indexes? O remember tx2 had something like that. @telphan ? @floion ?

floion: Yes @agherzan @imrehg 
The edison has a different layout and the 5th partition is not the state partition
And @telphan can chime in for the TX2 / TX1 (actually should chime in)

telphan: Yes, state partition is different in TX1 and TX2, I think it is 14 and 16 respectively.

agherzan: Thanks guys.

imrehg: @agherzan I have a better way to handle this in mind, will try it out and rewrite if that works.

imrehg:
@agherzan thanks! I'm checking with our solution architects, to see whether we'll need to merge this at all, or use the code separately (since it's only for one customer). But the feedback is very helpful so far!

imrehg:
@agherzan redone the state partition check, let me know if this is any better?
Also added everything that's needed to be able to update devices of this kind without any extra flags.
This kept as a single commit, as later we might remove these changes, if no longer needed.

xginn8:
closing for now, can reopen if necessary

PR #188, opened 2018-10-18
upgrade-1.x-to-2.x: get the supervisor version from the new root

imrehg:
Previously we got it from S3, but actually we can get it from the new OS, similarly to 2.x-2.x updates.
Still need some code cleanup, but it should now work with the new supervisor releases for the new logging system.

Change-type: patch


imrehg:
@resin-ci retest please

PR #214, opened 2019-03-11
Add CODEOWNERS

imrehg:
Change-type: patch


imrehg:
@samothx based on feedback on other threads, I've added a second code owner, so it's not just the person who adds the code usually who is the owner, but there's prefill for me as well if I push code.

Thanks!

PR #219, opened 2019-03-18
upgrade-2.x: user tempfile for FIFO

imrehg:
Using temp files instead of fixed file name on the host OS side fixes an issue when the update has been interrupted, but the OS hasn't been restarted before the next update attempt. In that case repeated `mkfifo` would fail with the same filename. Instead of reusing the filename, properly use a tempfile that is not clashing. Thus if the update fails, it can be safely restarted (and a new filename would be used).

Change-type: patch

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -359,6 +359,7 @@ function in_container_hostapp_update {
     local target_dockerd
     local volumes_args=()
     local retrycount=0
+    local tmp_image
 
     stop_services
     if [ "${STOP_ALL}" == "yes" ]; then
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -387,8 +388,11 @@ function in_container_hostapp_update {
             sleep 10
         fi
     done
-    mkfifo /tmp/resinos-image.docker
-    ${DOCKER_CMD} save "${update_package}" > /tmp/resinos-image.docker &
+
+    tmp_image=$(mktemp -u)
```
agherzan: Would be nice to have a prefix here to namespace these files.

imrehg: @agherzan added, see the latest version of the PR. Will add another PR for namespacing the other case where we use `mktemp` as well.

agherzan: Looks good.

PR #221, opened 2019-03-25
upgrade-2.x: trigger mounting the inactive root partition manually

imrehg:
From an upcoming OS release, that partition might not be mounted by default. Make sure that when it is needed, it is manually mounted nonetheless.

Connects-to: https://github.com/balena-os/meta-balena/pull/1441
Change-type: minor


imrehg:
Pending, because might not be required, if the OS doesn't implement the changes that necessitate this PR

PR #222, opened 2019-03-28
upgrade-2.x: redo supported device slug test

imrehg:
The test previously conflated two different things: device type support
and setting a variable for device architecture that is needed in one of
the hotpatch steps. Now disentangle these:

* make the hotpatch-related architecture check automatic, and do it when   it is needed
* only do device type support check if the update version is not to a   hostapp-enabled version
* make it clearer that the device support check is needed to know how   to switch the boot partition on those OS versions (when it's not   the hostapp update hooks that do that)
* deprecate the `--assume-supported` flag, as it has no meaning now, practically

Change-type: minor

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -110,10 +110,10 @@ Options:
         Use with extreme caution!
 
   --assume-supported
-        Assuming supported device, and disabling the relevant check.
-        Only enabled for updates that use update hooks, otherwise the updater
-        wouldn't know how to switch partitions, so only available for resinOS ${minimum_hostapp_target_version}.
-        You probably want to supply your own '--resinos-tag' as well in this case.
+        This is now deprecated. Assuming supported device, and disabling the relevant check.
+        Only enabled for updates that does not use update hooks, otherwise the updater
+        wouldn't know how to switch partitions, so only available for resinOS
+        below ${minimum_hostapp_target_version}.
 EOF
 }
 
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -815,16 +815,18 @@ fi
...
         binary_type=arm
         ;;
     intel-edison|intel-nuc|iot2000|up-board|qemux86*)
         binary_type=x86
         ;;
     *)
-        if { version_gt "$target_version" "$minimum_hostapp_target_version" || [ "$target_version" == "$minimum_hostapp_target_version" ]; } &&
+        if version_gt "${VERSION_ID}" "${minimum_hostapp_target_version}" || [ "${VERSION_ID}" == "${minimum_hostapp_target_version}" ] ; then
+            log "Update implicitly supported with hostapp-based updates for ${SLUG} device type."
+        elif { version_gt "$target_version" "$minimum_hostapp_target_version" || [ "$target_version" == "$minimum_hostapp_target_version" ]; } &&
```
xginn8: small nit: can we style the variables consistently here?
```suggestion
        elif { version_gt "${target_version}" "${minimum_hostapp_target_version}" || [ "${target_version}" == "${minimum_hostapp_target_version}" ]; } &&
```

imrehg: @xginn8 fair, should have done it already... Will do that. 
Also, these scripts would need some bit of housekeeping (of not real changes to functionality, just mostly cleaning up - hopefully some day soon)

imrehg: @xginn8 pushed changes as per your uggestion

imrehg:
@hedss no, devices show up as as `i586` (IOT2000)/`i686`(Edison), nothing below that. `uname -i` if works on the device, then it would give `i386` (not `uname -m` which gives these results). But repushed with a wildcard check of that for you :)

PR #226, opened 2019-04-22
hup-1.x: allow updater to take a v1 Docker registry URL

imrehg:
Instead of being hard-coded into the image.

Change-type: minor


imrehg:
And #227 completes this on the scripts side.

PR #233, opened 2019-04-24
upgrade-2.x: increase timeout values during update to 5 minutes

imrehg:
The previous timeout was 0.2s * 150=30s, which is got hit by some otherwise normally working devices, that might be a) slower than usual, b) the user application might be taking too much of the resources. Increase this x10, 0.2s * 1500=300s=5min, which is still quite short, and definitely shorter than if the user needs to ask support and retry because of "spurious" timeouts. Valid errors should be still caught quite quickly.

Change-type: patch

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -418,7 +418,7 @@ function in_container_hostapp_update {
       --privileged \
       "${volumes_args[@]}" \
       "${update_package}" \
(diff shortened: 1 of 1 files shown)


imrehg:
@agherzan yeah, was thinking about that, some of the escaping is a bit convoluted there, hence i didn't attempt it yet, but will revisit for a future PR. Thanks for the feedback :)

PR #236, opened 2019-05-08
upgrade-2.x: sanitize persistent logging device config vars

imrehg:
This is to fix a missmatch between supervisor's expectations between
versions (older adds the incorrect empty string, while newer borks
on that value).

Change-type: patch

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -235,6 +235,9 @@ function upgrade_supervisor() {
     else
         log WARN "Could not parse current supervisor version from the API, skipping update..."
     fi
+
+    # Post supervisor update fixes
+    persistent_logging_config_var
 }
 
 function error_handler() {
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -337,6 +340,38 @@ function pre_update_fix_bootfiles_hook {
...
+            "${API_ENDPOINT}/v5/device_config_variable(${PROBLEMATIC_ENV_VAR})" \
+            -H "Content-Type: application/json" \
+            -H "Authorization: Bearer ${APIKEY}" \
+            --data '{
+                "value": "false"
+            }' >> /dev/null
+        log "Updating config.json with sanitized '.persistentLogging' value."
+        tmpfile=$(mktemp -t configjson.XXXXXXXX)
+        jq '.persistentLogging="false"' < "${CONFIGJSON}" > "${tmpfile}"
+        cp "${tmpfile}" "$CONFIGJSON" && sync
```
agherzan: Shouldn't this be a rename (mv)?

imrehg: @agherzan much difference in this case? Since it's a copy anyways (from memory on `/tmp` to SD card on `/mnt/boot`), I think it always needs a copy.

imrehg: I've updated, but the question stands.

upgrade-2.x.sh
```diff
@@ -337,6 +340,38 @@ function pre_update_fix_bootfiles_hook {
     mount --bind "$bootfiles_temp"  /etc/hostapp-update-hooks.d/0-bootfiles
 }
 
+#######################################
+# Updat problematic persistent logging env var
```
CameronDiver:
```suggestion
# Update problematic persistent logging env var
```

imrehg: :P

upgrade-2.x.sh
agherzan: I was pointing at the atomicity of the operation. Missed the fact that they are on different fs but I think we need to make this operation atomic.

PR #240, opened 2019-06-19
logging: redirect ERROR messages to stderr (to pass back to the proxy)

xginn8:
Change-type: patch

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -134,10 +134,12 @@ function log {
             ;;
     esac
     endtime=$(date +%s)
-    printf "[%09d%s%s\n" "$((endtime - starttime))" "][$loglevel]" "$1"
     if [ "$loglevel" == "ERROR" ]; then
+        printf "[%09d%s%s\n" "$((endtime - starttime))" "][$loglevel]" "$1" >> /dev/stderr
         progress 100 "OS update failed"
         exit 1
+    else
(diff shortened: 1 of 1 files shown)


xginn8:
this change is a small nicety that i noticed while debugging something else entirely.

before: 
resinhup-no-stderr

after:
resinhup-with-stderr

xginn8:
It should be part of coreutils, @agherzan do you mean generally across OS
versions or specifically before calling the binary?

agherzan:
We introduce a tool dependency on the OS. How do we maintain that? We can't assume any dependency - not even coreutils. This is a question for @imrehg too.

xginn8:
@agherzan since this is provided as part of a busybox applet, i see [the common core busybox config](https://github.com/balena-os/meta-balena/blob/ce3efa3a66b91522d8beec6a4d9eb76c5b73a5e6/meta-balena-common/recipes-core/busybox/busybox/defconfig#L294) contains the requisite config. I also see [rocko](https://github.com/balena-os/meta-balena/blob/609118963ab642a858469202d36538a7978fa1b2/meta-resin-rocko/recipes-core/busybox/busybox/defconfig#L294) and [pyro](https://github.com/balena-os/meta-balena/blob/609118963ab642a858469202d36538a7978fa1b2/meta-resin-pyro/recipes-core/busybox/busybox/defconfig#L294) having this config variable defined, do you have a way of tracking what depends on these various busybox configuration flags?

agherzan:
That's the food for thought here. We don't have a way now to express this dependency.

PR #247, opened 2019-08-08
replace "nohup ..." with "systemd reboot" in reboot requests

### LINKED ISSUE #245: Use systemd run instead of nohup for reboot
Better way to do temporary scripts

imrehg:
Closes #245

Change-type: patch

### DIFF
--- a/upgrade-1.x-to-2.x.sh
+++ b/upgrade-1.x-to-2.x.sh
@@ -990,7 +990,16 @@ if [ "$NOREBOOT" == "no" ]; then
     # Reboot into new OS
     log "Rebooting into new OS in 5 seconds..."
     progress 100 "Update successful, rebooting"
-    nohup bash -c " /bin/sleep 5 ; /sbin/reboot " > /dev/null 2>&1 &
+    systemd-run --on-active=5 --unit=hup-reboot.service systemctl reboot
+    sleep 300
+    # If the previous reboot command has failed for any reason, let's try differently
+    nohup bash -c "reboot --force" > /dev/null 2>&1 &
+    sleep 300
+    # If the previous 2 reboot commands have failed for any reason, try the Magic SysRq
+    # enable it
+    echo 1 > /proc/sys/kernel/sysrq
+    # send reboot request
+    echo b > /proc/sysrq-trigger
 else
     log "Finished update, not rebooting as requested."
     log "NOTE: Supervisor and stopped services kept stopped!"
--- a/upgrade-1.x.sh
+++ b/upgrade-1.x.sh
@@ -786,7 +786,16 @@ if [ $RESINHUP_EXIT -eq 0 ] || [ $RESINHUP_EXIT -eq 2 ] || [ $RESINHUP_EXIT -eq 
     # Everything is fine - Reboot
     if [ "$NOREBOOT" == "no" ]; then
         log "Rebooting board in 5 seconds..."
-        nohup bash -c " /bin/sleep 5 ; /sbin/reboot " > /dev/null 2>&1 &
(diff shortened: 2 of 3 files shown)


upgrade-1.x-to-2.x.sh
```diff
@@ -988,9 +988,9 @@ rm -rf /mnt/data/resin-data/resin-supervisor
 sync
 if [ "$NOREBOOT" == "no" ]; then
     # Reboot into new OS
-    log "Rebooting into new OS in 5 seconds..."
+    log "Rebooting into new OS..."
     progress 100 "Update successful, rebooting"
```
xginn8: Was the sleep intended to ensure the progress message makes it out before the reboot? If so, it may be worth prepending the reboot with the same `sleep 5;`

imrehg:
@agherzan it does use systemd instead of going around it, and a less tricky way, than `nohup`. `systemd reboot` is async, thus I think it would let things get out on time. Also asked for help on this at https://www.flowdock.com/app/rulemotion/r-resinos/threads/rWUM2bcrvUvBPzUFKxsJKG_a66r  and got some feedback from @xginn8 and @ZubairLK so far. The alternative that I was thinking about there is:
```
systemd-run --on-active=5 --unit=hup-reboot.service /sbin/reboot
```
but that's kinda mixing things>
I wonder how horrible it would be to switch to:
```
systemd-run --on-active=5 --unit=hup-reboot.service systemctl reboot
```
in which case we use systemd but also retain the delay.

agherzan:
@imrehg I agree that it's async but there are no guarantees that the next API calls (for progress) manage to finish in time. The 5 seconds delay was always enough for that. I do agree that the sleep implementation is not the best but it does what it is supposed to do while with this we risk breaking it. I was also thinking about `systemd-run`. That sounds a good way forward to me.

upgrade-1.x-to-2.x.sh
imrehg: @xginn8 done with a `systemd-run` native way.

imrehg:
Discussed more in https://www.flowdock.com/app/rulemotion/public-s-premium/threads/9BHjK1ooYRkjNIb45p1lwy3PMe0 to be even more robust, hopefully.

upgrade-1.x-to-2.x.sh
```diff
@@ -990,7 +990,16 @@ if [ "$NOREBOOT" == "no" ]; then
     # Reboot into new OS
     log "Rebooting into new OS in 5 seconds..."
     progress 100 "Update successful, rebooting"
-    nohup bash -c " /bin/sleep 5 ; /sbin/reboot " > /dev/null 2>&1 &
+    systemd-run --on-active=5 --unit=hup-reboot.service systemctl reboot
+    sleep 300
+    # If the previous reboot command has failed for any reason, let's try differently
+    nohup bash -c "reboot --force" > /dev/null 2>&1 &
```
xginn8: why not schedule this as a unit as well? does that somehow block the reboot process?

imrehg: @xginn8 if you check the thread with @ZubairLK , there the idea was: "if a systemctl reboot fails, [we might assume] we can't rely on systemd anymore."

imrehg: But yeah, doing the same as the first step just with `--force` or `--force --force` was also on the list. This is more "defence in depth", I guess.

imrehg:
Since this form is as @ZubairLK proposed in chat, and it's strictly speaking as good or better than the previous version, I'll going to merge it, so other PRs can move forward.

PR #248, opened 2019-08-28
Remove "--nolog" flag from 2.x updater, and always set up logfile

imrehg:
Previously we might have logged when there wasn't any logfile setup, for example if any of the argument parsing was throwing an error, and on the 2.x updater there wasn't any way to directly know what the error was, just that the update didn't start.

Moving the logfile setup before any logging takes place eliminates this issue. The flag was also only really used in manual HUP runs (no way to use in self-serve updates), and it's more relevant to 1.x updates than anything since.

Change-type: minor

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -2,7 +2,6 @@
 
 # default configuration
 NOREBOOT=no
-LOG=yes
 IGNORE_SANITY_CHECKS=no
 RESINOS_REGISTRY="registry.hub.docker.com"
 RESINOS_REPO="resin/resinos"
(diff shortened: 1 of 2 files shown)


xginn8:
Since this logging occurs on-disk, I wonder if it makes sense to create a file in tmpfs and then rotate it to disk.

imrehg:
@xginn8 to understand better, rotate when the update script exits? Or have something else in mind? Because in that case power cut/reboot would leave no trace of any ongoing HUP, that might be used to diagnose what happened.

xginn8:
Yeah, fair critique, disregard my comment!

PR #258, opened 2020-03-19
enable delta host OS upgrades

xginn8:
Depends-on: https://github.com/balena-io/balena-api/pull/2513
Depends-on: https://github.com/balena-io/balena-api/pull/2514
HQ: https://github.com/balena-io/balena/issues/1113
Change-type: minor

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -2,6 +2,7 @@
 
 # default configuration
 NOREBOOT=no
+DELTA_VERSION=3
 IGNORE_SANITY_CHECKS=no
 RESINOS_REGISTRY="registry.hub.docker.com"
 RESINOS_REPO="resin/resinos"
@@ -74,7 +75,7 @@ Options:
         If not defined, then the update will try to run for the HOSTOS_VERSION's
         original supervisor release.
 
-    -n, --nolog
+  -n, --nolog
         By default tool logs to stdout and file. This flag deactivates log to file.
 
   --no-reboot
@@ -420,8 +421,12 @@ function in_container_hostapp_update {
     fi
 
     while true ; do
-        ${DOCKER_CMD} pull "${update_package}" && break || log WARN "Couldn't pull docker image, was try #${retrycount}..."
-        retrycount=$[$retrycount+1]
+        if ${DOCKER_CMD} pull "${update_package}"; then
+            break
+        else
+            log WARN "Couldn't pull docker image, was try #${retrycount}..."
+        fi
+        retrycount+=1
         if [ $retrycount -ge 10 ]; then
             log ERROR "Couldn't pull docker image, giving up..."
         else
@@ -579,7 +584,7 @@ function hostapp_based_update {
             remove_containers
         fi
         log "Starting hostapp-update"
-        hostapp-update -i "${update_package}" || log ERROR "hostapp-update has failed..."-
+        hostapp-update -i "${update_package}"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


xginn8:
I'm happy with this as-is, though it requires API-side changes to allow devices to read and create deltas for public apps.

upgrade-2.x.sh
```diff
@@ -74,7 +75,7 @@ Options:
         If not defined, then the update will try to run for the HOSTOS_VERSION's
         original supervisor release.
 
-    -n, --nolog
+  -n, --nolog
```
xginn8, line 78: this is just for visual consistency

upgrade-2.x.sh
```diff
@@ -420,8 +421,12 @@ function in_container_hostapp_update {
     fi
 
     while true ; do
-        ${DOCKER_CMD} pull "${update_package}" && break || log WARN "Couldn't pull docker image, was try #${retrycount}..."
-        retrycount=$[$retrycount+1]
+        if ${DOCKER_CMD} pull "${update_package}"; then
```
xginn8, line 424: these changes are to appease shellcheck

saintaardvark: :+1: I'm a fan of the more explicit notation too.

upgrade-2.x.sh
```diff
@@ -940,12 +1001,20 @@ target_version=$(echo "$target_version" | tr + _)
 if [ -z "$RESINOS_TAG" ]; then
     RESINOS_TAG=${target_version}-${SLUG}
 fi
-image="${RESINOS_REGISTRY}/${RESINOS_REPO}:${RESINOS_TAG}"
-log "Checking for manifest of ${image}"
-if [ "$(image_exists "$RESINOS_REGISTRY" "$RESINOS_REPO" "$RESINOS_TAG")" = "yes" ]; then
-    log "Manifest found, good to go..."
+log "Attempting host OS update using deltas"
+
+image=$(find_delta "${target_version/_/+}")
```
saintaardvark, line 1006: It looks like we're normalizing `$target_version` here, but I don't know the reason.  Is it appropriate to add a comment?

If it is being normalized for use by `find_delta()`, should it be moved into that function along with the other normalization being done thee? It looks like this is the only place in the script where we're making this change.

upgrade-2.x.sh
```diff
@@ -940,12 +1001,20 @@ target_version=$(echo "$target_version" | tr + _)
...
+if [ -z "${image}" ]; then
```
saintaardvark, line 1007: I might suggest restructuring this a bit:

```
if [ "${image}" ] ; then
  log "Found delta located at ${image}"
else
  ...
```

This makes for a shorter (visual) distance to keep the context of the check, compared to having it after a failure check with its own nested check.

xginn8: Yep, that makes sense, I'll flip this logic. Good note!

upgrade-2.x.sh
xginn8: This can be cleaned up further to not require the substitution, have pushed up a cleaner version.

upgrade-2.x.sh
```diff
@@ -696,6 +701,61 @@ function find_partitions {
...
+# Returns:
+#   Registry URL for desired image
+#######################################
+function get_image_location() {
+    local version=$1
+    # TODO: could improve the quality of the API call here
+    curl --retry 10 --silent -X GET \
+        -H "Content-Type: application/json" \
+        -H "Authorization: Bearer ${APIKEY}" \
+        "${API_ENDPOINT}/v5/release?\$expand=release_tag,belongs_to__application,contains__image/image&\$filter=(belongs_to__application/any(a:a/device_type%20eq%20'${SLUG}'))%20and%20(release_tag/any(rt:(rt/tag_key%20eq%20'version')%20and%20(rt/value%20eq%20'${version}')))" \
```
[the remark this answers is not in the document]
xginn8: @Page- is the odata query guru and can probably make this a more reasonable query. This script has a few complex odata queries, but is probably one of the only places in the host OS that we're manipulating anything complex. I don't know that the juice is worth the squeeze (especially as this codepath should disappear before long anyway once the supervisor manages host app updates as well).

xginn8:
> As far as I can tell, this looks good to me. A couple of reservations:
> 
>     * I'm unable to verify the API calls to the Delta endpoint; is there documentation for this that I've missed?

I'm not sure if/where this is documented, but I used the supervisor's source to generate this call: https://github.com/balena-io/balena-supervisor/blob/d5767b647d429540abf34e28ad26c7ca59688f52/src/lib/docker-utils.ts#L132.

> 
>     * As mentioned, the API query in `get_image_location()` is complex.  I'm unsure if it's feasible to consider breaking up the building of that query to smaller chunks.

upgrade-2.x.sh
```diff
@@ -2,6 +2,7 @@
 
 # default configuration
 NOREBOOT=no
+DELTA_VERSION=3
```
xginn8, line 5: I've tested this back to the earliest multicontainer-capable OS version (2.11.0), but it'd be good to actually confirm these deltas can be applied before that. @robertgzr do you know when balena-engine deltas became fully supported in the OS?

upgrade-2.x.sh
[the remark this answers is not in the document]
saintaardvark: Gotcha.  If it's going to disappear shortly, then I'm happy to leave this for now.

upgrade-2.x.sh
dfunckt: I don’t remember the exact version off the top of my head, I’ll check tomorrow, but they’re generally available since the OS version that switched to balenaEngine from docker.

upgrade-2.x.sh
```diff
@@ -696,6 +701,61 @@ function find_partitions {
...
+    local src_image target_image
+    # shellcheck disable=SC2153
+    src_image=$(get_image_location "${VERSION}")
+    target_image=$(get_image_location "${target_version}")
+    if [ -z "${src_image}" ] || [ -z "${target_image}" ]; then
+        return
+    else
+        # TODO: should we retry this more extensively? deltas may take a while to generate..
+        delta=$(curl --retry 10 --silent -X GET \
+            "${DELTA_ENDPOINT}/api/v${DELTA_VERSION}/delta?src=${src_image}&dest=${target_image}" \
```
dfunckt, line 750: This needs to be retrying significantly more than 10 times, deltas can take an arbitrary amount of time to be generated on demand, typically depending on the sizes of the images. You’ll also ideally backoff a bit. Lastly, for any response code other than 200 and 503, you’ll need to stop polling and either fail the script or skip deltas.

On the other hand, it’s going to be far better if deltas are pregenerated when the OS releases are made, so this call returns immediately.

upgrade-2.x.sh
dfunckt: Okay, found it, the oldest available OS version with balenaEngine is v2.9.0: https://github.com/balena-os/meta-balena/blob/master/CHANGELOG.md#v290

xginn8: Awesome, thanks for confirming @dfunckt. I've confirmed a few different upgrade paths around that changeover, so looks good.

upgrade-2.x.sh
xginn8: There's an open question on whether or not we want to pregenerate the deltas -- probably best to discuss in the mini-spec (https://github.com/balena-io/balena/pull/2127)?

xginn8:
@saintaardvark @ZubairLK @Page- PTAL as this code as-is is deployable, despite the API side permissions not being in place. Since we fall back to a regular pull, as-is this'll only add a single API call to what's done already. I'm working on the permissioning separately.

xginn8:
> Any chance we could migrate to naming it as `balenahup`?

Renaming the repository should be straightforward. There are lots of other `resin` references to clean up as well, and I'm not quite sure how much is worth cleaning up (assuming this'll be deco'd entirely within a ~year). I could do a cosmetic pass, but since lots of tools and bits cross-reference meta-balena I'm reticent to do too much for fear of breaking some ancient update codepath.

PR #259, opened 2020-04-28
TLS: reuse balenaRootCA to bypass local MitM and self-signed certs

xginn8:
Change-type: patch

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -46,6 +46,11 @@ _prepare_locking()  { eval "exec $LOCKFD>\"$LOCKFILE\""; trap _no_more_locking E
 # Public functions
 exlock_now()        { _lock xn; }  # obtain an exclusive lock immediately or fail
 
+# workaround for self-signed certs, waiting for https://github.com/balena-os/meta-balena/issues/1398
+TMPCRT=$(mktemp)
+jq -r '.balenaRootCA' < /mnt/boot/config.json | base64 -d > "${TMPCRT}"
+cat /etc/ssl/certs/ca-certificates.crt >> "${TMPCRT}"
+
 # Dashboard progress helper
 function progress {
     percentage=$1
@@ -206,7 +211,7 @@ function upgrade_supervisor() {
         fi
     fi
 
-    if CURRENT_SUPERVISOR_VERSION=$(curl --retry 10 --silent --header "Authorization: Bearer ${APIKEY}" "${API_ENDPOINT}/v5/device(${DEVICEID})?\$select=supervisor_version" | jq -r '.d[0].supervisor_version'); then
+    if CURRENT_SUPERVISOR_VERSION=$(CURL_CA_BUNDLE=${TMPCRT} curl --retry 10 --silent --header "Authorization: Bearer ${APIKEY}" "${API_ENDPOINT}/v5/device(${DEVICEID})?\$select=supervisor_version" | jq -r '.d[0].supervisor_version'); then
         if [ -z "$CURRENT_SUPERVISOR_VERSION" ]; then
             log ERROR "Could not get current supervisor version from the API..."
         else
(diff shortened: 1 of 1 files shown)


saintaardvark:
Hey @xginn8  - I'm taking a look at this now, but are you able to paste any testing output in here? It would be good to see some output from this, even if it's just so I can learn how to do this kind of testing properly.

upgrade-2.x.sh
```diff
@@ -205,18 +210,18 @@ function upgrade_supervisor() {
         fi
     fi
 
-    if CURRENT_SUPERVISOR_VERSION=$(curl --retry 10 --silent --header "Authorization: Bearer ${APIKEY}" "${API_ENDPOINT}/v5/device(${DEVICEID})?\$select=supervisor_version" | jq -r '.d[0].supervisor_version'); then
+    if CURRENT_SUPERVISOR_VERSION=$(CURL_CA_BUNDLE=${TMPCRT} curl --retry 10 --silent --header "Authorization: Bearer ${APIKEY}" "${API_ENDPOINT}/v5/device(${DEVICEID})?\$select=supervisor_version" | jq -r '.d[0].supervisor_version'); then
```
saintaardvark, line 213: Is it worth breaking out the curl commands to a separate command?  Maybe adding something after the setup of the TMPCRT file:

```
CURLB="CURL_CA_BUNDLE=${TMPCRT} curl"
```

upgrade-2.x.sh
```diff
@@ -336,7 +341,7 @@ function pre_update_fix_bootfiles_hook {
     log "Applying bootfiles hostapp-hook fix"
     local bootfiles_temp
     bootfiles_temp=$(mktemp)
-    curl --retry 10 -f -s -L -o "$bootfiles_temp" https://raw.githubusercontent.com/resin-os/resinhup/77401f3ecdeddaac843b26827f0a44d3b044efdd/upgrade-patches/0-bootfiles || log ERROR "Couldn't download fixed '0-bootfiles', aborting."
+    CURL_CA_BUNDLE=${TMPCRT} curl --retry 10 -f -s -L -o "$bootfiles_temp" https://raw.githubusercontent.com/resin-os/resinhup/77401f3ecdeddaac843b26827f0a44d3b044efdd/upgrade-patches/0-bootfiles || log ERROR "Couldn't download fixed '0-bootfiles', aborting."
```
saintaardvark, line 344: This line (and some that follow) use shorter flags (`-s` vs the `--silent` earlier in the script).  I'm a fan of the more explicit flags, but it's probably worth consistifying things here one way or the other (or possibly putting them into the `CURLB` variable I suggested earlier, if there's enough common ground).

upgrade-2.x.sh
```diff
@@ -976,7 +982,7 @@ if ! version_gt "$VERSION" "$preferred_hostos_version" &&
             download_uri=https://github.com/resin-os/resinhup/raw/master/upgrade-binaries/$binary_type
             for binary in $tools_binaries; do
                 log "Installing $binary..."
-                curl --retry 10 -f -s -L -o $tools_path/$binary $download_uri/$binary || log ERROR "Couldn't download tool from $download_uri/$binary, aborting."
+                CURL_CA_BUNDLE=${TMPCRT} curl --retry 10 -f -s -L -o $tools_path/$binary $download_uri/$binary || log ERROR "Couldn't download tool from $download_uri/$binary, aborting."
```
saintaardvark, line 985: Here and up ahead, should this line have dis-ambiguating curly braces around the variables?  Ie:

```
${tools_path}/${binary} ${download_uri}/${binary}
```

upgrade-2.x.sh
```diff
@@ -998,7 +1004,7 @@ if version_gt "$VERSION_ID" "2.0.6" &&
         mkdir -p $tools_path
         export PATH=$tools_path:$PATH
         download_url=https://raw.githubusercontent.com/resin-os/meta-resin/v2.3.0/meta-resin-common/recipes-support/resin-device-progress/resin-device-progress/resin-device-progress
-        curl --retry 10 -f -s -L -o $tools_path/resin-device-progress $download_url || log WARN "Couldn't download tool from $download_url, progress bar won't work, but not aborting..."
+        CURL_CA_BUNDLE=${TMPCRT} curl --retry 10 -f -s -L -o $tools_path/resin-device-progress $download_url || log WARN "Couldn't download tool from $download_url, progress bar won't work, but not aborting..."
```
saintaardvark, line 1007: There are a couple of places in the existing code where we're hardcoding the output directory -- [/tmp/upgrade_tools](https://github.com/balena-os/resinhup/blob/master/upgrade-2.x.sh#L956) and [/tmp/upgrade_tools_extra](https://github.com/balena-os/resinhup/blob/master/upgrade-2.x.sh#L997).  It would be worth using `mktemp`, as we do up above for the curl CA bundle.

xginn8:
> Hey @xginn8 - I'm taking a look at this now, but are you able to paste any testing output in here? It would be good to see some output from this, even if it's just so I can learn how to do this kind of testing properly. grin

There are a few ways I've tested this -- first, locally on my BoB (in fact, without this change HUP will not work locally). Debug log attached:
[hup-bob-debug.txt](https://github.com/balena-os/resinhup/files/4548522/hup-bob-debug.txt)

Second, I copied the script to a device on staging and called it locally in debug with the args it expects:

[staginghup.txt](https://github.com/balena-os/resinhup/files/4548561/staginghup.txt)

I have also tested an old rev to the next rev on staging (2.2.0+rev1 -> 2.2.0+rev2) in the same way. 
[old-hup.txt](https://github.com/balena-os/resinhup/files/4548707/old-hup.txt)


Once this code lands on staging we can run the hup-ladder (https://github.com/balena-io-playground/hup-ladder).

saintaardvark:
Awesome, @xginn8!  Thanks very kindly for that - I'm taking a look now & am also aiming to duplicate https://github.com/balena-io/device-diagnostics/pull/199/ on my BoB. Testing this with the hup-ladder is something I had not thought about -- good point.

upgrade-2.x.sh
xginn8: I don't like changing things like this out-of-scope, but I agree it could be standardized. There's no real difference unless you're concatenating strings (i.e. `${a_string}another_string${a_third_string}` != `$a_stringanother_string$a_third_string`), so I'd prefer to leave it for now and clean all those up in one fell swoop.

upgrade-2.x.sh
xginn8: A fair point indeed, again I'd rather change this on it's own. These dirs predate my involvement with this process, but I suspect they aren't mktemps to make future debugging easier to reason about in the case that they failed, but just speculation.

xginn8:
@balena-ci retest

xginn8:
This is still not 100% sufficient to make HUP work flawlessly, as progress is reported by a script on-device called `resin-device-progress` and that script still does not honor these bundles. Same goes for `update-resin-supervisor`.

xginn8:
@balena-ci rebase

PR #260, opened 2020-04-30
hostapp-images: try pulling from balena registry before DockerHub

xginn8:
Change-type: minor

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -714,7 +714,8 @@ function find_partitions {
 #   Registry URL for desired image
 #######################################
 function get_image_location() {
-    local version=$1
+    # we need to strip the target_version's variant tag to query the API properly
+    local version=${1/.dev/}
     # TODO: could improve the quality of the API call here
     curl --retry 10 --silent -X GET \
         -H "Content-Type: application/json" \
@@ -759,8 +760,7 @@ function get_delta_token() {
 #   Location of delta image
 #######################################
 function find_delta() {
-    # we need to strip the target_version's variant tag to query the API properly
-    local target_version=${1/.dev/}
+    local target_version=${1}
     local src_image target_image
     # shellcheck disable=SC2153
     src_image=$(get_image_location "${VERSION}")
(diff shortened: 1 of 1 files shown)


xginn8:
opening this so i don't forget about it (cc @shaunmulligan)

upgrade-2.x.sh
```diff
@@ -760,7 +760,7 @@ function get_delta_token() {
 #######################################
 function find_delta() {
     # we need to strip the target_version's variant tag to query the API properly
```
saintaardvark, line 762: - Should this comment be removed, or possibly moved up to `get_image_location()`?

- Out of curiosity, what prompted the change?

upgrade-2.x.sh
```diff
@@ -1178,12 +1173,21 @@ if version_gt "${VERSION_ID}" "${minimum_hostapp_target_version}" ||
...
     log "hostapp-update command exists, use that for update"
     progress 50 "Running OS update"
-    if [ -n "${delta_image}" ] && hostapp_based_update "${delta_image}"; then
-        # overriding this for the later supervisor version inspection, since we've updated successfully
-        image=${delta_image}
-    else
-        log "Delta failed or not found, falling back to regular pull"
-        hostapp_based_update "${image}" || log ERROR "hostapp-update has failed..."
+    images=("${delta_image}" "$(get_image_location "${target_version/_/+}")" "${upstream_image}")
+    image_types=("delta" "balena_registry" "upstream_registry")
```
saintaardvark, line 1177: It looks like you're essentially building an associative array.  Would it be worth using that Bash feature directly?  I'm not a fan of the Bash syntax for arrays (linear or associative), so if you want to skip it, it would be worth adding a comment to make the purpose clear.

upgrade-2.x.sh
xginn8: good catch! I moved it since i call that same function a second time in line 1176, and it's a transformation that more precisely should be done there anyway since that's where it's used (in `get_image_location()`).

upgrade-2.x.sh
xginn8: Shall add a comment, i really don't care for the associative arrays in bash!

PR #262, opened 2020-05-04
delta-hup: include size of delta

xginn8:
Connects-to: https://github.com/balena-os/resinhup/issues/261
Change-type: patch

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -1034,7 +1034,11 @@ else
     log "Device not delta capable"
 fi
 if [ -n "${delta_image}" ]; then
-    log "Found delta image: ${delta_image}"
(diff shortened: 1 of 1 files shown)


xginn8:
> This looks good to me, but is there a way to test this? It's low-consequences if it's incorrect, but it would be good to know that this can be tested in advance.

you can test on BoB (simply build the proxy from source with this file dropped in). i will also sometimes copy the file to disk (`/tmp`) and run it locally while developing.

From my BoB:
```
[000000012][LOG]Found delta image: registry2.my.devenv.local/v2/e6a131896860245cf14eabeb3f3d603d:delta-095d46d1470fc0aa, size: 16 MB
```

PR #286, opened 2020-06-23
use device uuid not numeric id

xginn8:
HQ: https://github.com/balena-io/balena-io/issues/2227
Change-type: patch


saintaardvark:
This looks good to me, but what's the advantage of using UUID instead of the numeric ID?

xginn8:
@saintaardvark there's an FD conversation in the linked HQ ticket that details the failure mode.

PR #287, opened 2020-06-24
use device uuid not numeric id in legacy scripts

xginn8:
HQ: https://github.com/balena-io/balena-io/issues/2227
Change-type: patch


xginn8:
@balena-ci I self-certify!

PR #290, opened 2020-07-17
non-hostapp: fix up retry count non-integer issue

xginn8:
since the variable is not set as an integer, we are simply appending to
the string rather than incrementing the value:

```
[000000009][WARNING]Couldn't pull docker image, was try #0...
[000000019][WARNING]Couldn't pull docker image, was try #01...
```

Change-type: patch


xginn8:
@balena-ci I self-certify!

xginn8:
Relates to https://github.com/balena-os/balenahup/issues/291

Issue #296, opened 2020-08-05
Migrate `REGISTRY_ENDPOINT` from locally sourced to called as an arg


xginn8:
ideally we shouldn't use `REGISTRY_ENDPOINT` [from the device directly in HUP] though because it's likely to break.

we can pass this in as an argument from the proxy instead

internal ref. https://www.flowdock.com/app/rulemotion/resin-starters/threads/JAAWksJVigHvllRXRGAfUc74RXI

xginn8:
We're already passing in these args to the HUP command anyway, see https://github.com/balena-io/balena-proxy/blob/master/src/services/actions-backend/app.ts#L190-L199. We'll need to update the flux deploy https://github.com/balena-io/environment-production/blob/8ad0ad03e0b62919f5f8cb8e3ff56a4c8e1b361f/kubernetes/balenacloud/balena-proxy/configmap.yaml#L8 and https://github.com/balena-io/environment-staging/blob/60b787eed04f4c690207d70d4a3689c6e3e9a8ba/kubernetes/balenacloud/balena-proxy/configmap.yaml#L8

xginn8:
This work is now complete & deployed in production.

PR #297, opened 2020-08-05
registry: use provided arg for FQDN rather than local source

xginn8:
Connects-to: https://github.com/balena-os/balenahup/issues/296
Change-type: patch


xginn8:
We're already passing in these args to the HUP command anyway, see https://github.com/balena-io/balena-proxy/blob/master/src/services/actions-backend/app.ts#L190-L199.

PR #303, opened 2020-09-18
jetson-tx2: fix for accidental overwrite of extlinux.conf

xginn8:
As sorted in https://github.com/balena-os/meta-balena/pull/2013, we
missed some files in the hooks blacklist that cause devices to brick in
the field. until all jetson-tx2 devices are past balenaOS 2.58.3, we'll
need to sort this as part of the HUP process.
    
Change-type: patch

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -304,12 +304,44 @@ function pre_update_fix_bootfiles_hook {
 
 function pre_update_jetson_fix {
     log "Caching current extlinux.conf for ${SLUG} fix"
-    cp /mnt/boot/extlinux/extlinux.conf /tmp/extlinux.conf
+    extlinux_path=/boot/extlinux
+    mkdir -p "/tmp${extlinux_path}"
+    cp "/mnt${extlinux_path}/extlinux.conf" "/tmp${extlinux_path}/extlinux.conf"
+}
+
+function parse_isolcpus {
+    path=$1
+    if grep -q "isolcpus=" "${path}" ; then
+        # shellcheck disable=SC2013
+        for val in $(awk '/isolcpus=/' "${path}"); do
+            if echo "${val}" | grep -q "isolcpus="; then
+                echo "${val}"
+            fi
+        done
+    fi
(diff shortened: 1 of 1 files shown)


xginn8:
We need to confirm the minimum OS version required as part of this workaround, cc @szony

upgrade-2.x.sh
```diff
@@ -302,6 +302,50 @@ function pre_update_fix_bootfiles_hook {
     mount --bind "$bootfiles_temp"  /etc/hostapp-update-hooks.d/0-bootfiles
 }
 
+function pre_update_jetson_fix {
+    log "Caching current extlinux.conf for ${SLUG} fix"
+    extlinux_path=/boot/extlinux
```
shaunmulligan, line 307: we should ensure that `/boot/extlinux` is the exactly correct location. I know there was some movement of this file from boot to rootfs or something, so worth checking with @acostach the exact l location this file needs to end up in the 2.45.1 OS version to ensure it works

upgrade-2.x.sh
```diff
@@ -302,6 +302,50 @@ function pre_update_fix_bootfiles_hook {
...
+                echo "${val}"
+            fi
+        done
+    fi
+}
+
+function post_update_jetson_fix {
+    log "Applying extlinux.conf fix for ${SLUG}"
+    # check if current config has isolcpus set in extlinux.conf
+    extlinux_file=/boot/extlinux/extlinux.conf
```
shaunmulligan, line 329: sorry my previous comment actually probably applies to this location.

alexgg (approved):
Looks good.

upgrade-2.x.sh
xginn8: this is a little unclear in the implementation, but it's actually targeting `/mnt${extlinux_file}` (aka `/mnt/boot/extlinux/extlinux.conf`). I'm updating the code to make this clearer, thanks for pointing this out.

xginn8:
@balena-ci I self-certify!

PR #306, opened 2020-09-21
post-update-workarounds: allow `hostapp-update` to fail properly

xginn8:
In order to detect issues with deltas and work around them with a
regular pull, we need `hostapp-update` to fail and propagate that back
to our caller. By splitting out these post-update steps into a different
function, we can `&&` them together and restore the previous behavior.

Change-type: patch


xginn8:
@balena-ci I self-certify!

PR #311, opened 2020-10-15
Remove device sanity checks, as they are no longer needed

saintaardvark:
This PR removes the sanity check in balenahup that verifies the device type.  This check examined two on-device files, in an attempt to prevent HUPs from installing OS versions incompatible with the underlying hardware.

The prompt for this change is to enable hands-off device-type changes for customers with private device types; by
removing the check, we can patch the device type in the API, and leave OS upgrades to the customer.

This has been checked on devices in staging by invoking it manually against my test devices.  In both the usual case (device type requested matches `device-type.json`) and the unusual case (device type requested does *not* match `device-type.json`), the upgrade proceeds as expected (the OS for the requested device type is fetched).

This has *not* been tested on my devenv due to https://github.com/balena-io/balena-on-balena/issues/193.  Because of that, I'm doing to mark this as draft for now until everyone has had a chance to review.

Change-type: patch
Signed-off-by: Hugh Brown <hugh@balena.io>

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -3,7 +3,6 @@
 # default configuration
 NOREBOOT=no
 DELTA_VERSION=3
-IGNORE_SANITY_CHECKS=no
 SCRIPTNAME=upgrade-2.x.sh
 LEGACY_UPDATE=no
 STOP_ALL=no
@@ -100,13 +99,6 @@ Options:
   --stop-all
         Request the updater to stop all containers (including user application)
         before the update.
-
-  --ignore-sanity-checks
-        The update scripts runs a number of sanity checks on the device, whether or not
-        it is safe to update (e.g. device type and running system cross checks)
-        This flags turns sanity check failures from errors into warnings only, so the
-        the update is not stopped if there are any failures.
-        Use with extreme caution!
 
   --assume-supported
         This is now deprecated. Assuming supported device, and disabling the relevant check.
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -998,19 +987,6 @@ UUID=$(jq -r '.uuid' $CONFIGJSON)
 API_ENDPOINT=$(jq -r '.apiEndpoint' $CONFIGJSON)
 DELTA_ENDPOINT=$(jq -r '.deltaEndpoint' $CONFIGJSON)
 
-## Sanity checks
-device_type_check=$(device_type_match)
```
sradevski, line 1002: Let's also remove the function itself

xginn8:
in my notes i have the following as a refactored sanity check, rather than a removed one:

> hup check should be new API-device type and local device-type.json

is there a reason we're not doing that as agreed upon?

sradevski:
@xginn8 We were trying to recall what was said in the call, since we didn't record it, unfortunately. I remembered the part where we said the current check is pretty useless in today's era, but ti seems I missed that we said we need to change it so it checks against the API's DT. 

We can maybe chat further on how to handle the DT migrations while having a DT check, since that will prevent self-served HUPs. Maybe we get on a call later today?

xginn8:
> We can maybe chat further on how to handle the DT migrations while having a DT check, since that will prevent self-served HUPs.

iiuic, masquerading devices should match between those two sources of truth (`device-type.json` and the migrated API DT). is that not correct? if anything, this'd prevent HUPs for masquerading devices that have not yet been migrated, which might not be a bad thing.

sradevski:
@xginn8 yeah you are right for some of the device types, I think the issue is for device types that we also renamed as part of adding them to cloud, which is 4 DTs ([here is the link](https://docs.google.com/spreadsheets/d/1qh_znN47ecDQwTxKEaGF1ielO_CS2H3eCg-GxsH9H2U/edit#gid=0)). I do see the point of having the check though, so maybe we'll have to do an on-device operation to change the slug on the device for these? Let's do a sync just so we are sure we have a clear migration process (maybe after the all-hands)?

sradevski (approved):

sradevski:
@balena-ci rebase

Issue #305, opened 2020-09-21
Handle migrating special device config for jetson-tx2


xginn8:
Some futureproofing is needed where we copy the config of `extlinux.conf` to `extra_uenv.txt` since that is where the supervisor will look from OS version 2.56. Additionally, there is ODMDATA configuration is stored directly in `/dev/mmcblk0boot0` partition, embedded in the signed booltloader bundle. This one is of course not a file in the boot partition to be blacklisted. The supervisor changes 1 byte in `/dev/mmcblk0boot0` to toggle between 6 settings. When performing HUP, the default byte will be updated in that partition, and if the supervisor detects the change, we will be prone to failure again.

ref. https://www.flowdock.com/app/rulemotion/r-resinos/threads/u8nVv8iy0CygFqemmhpOKLvu8I7
https://www.flowdock.com/app/rulemotion/resin-devices/threads/h4rhav-nPtFC6vROZyub47lVoRB

PR #312, opened 2020-10-16
jetson-tx2: migrate other special-case configuration during HUP

xginn8:
Change-type: patch
Connects-to: https://github.com/balena-os/balenahup/issues/305


saintaardvark (approved):

xginn8:
@balena-ci retest

PR #313, opened 2020-10-22
Make HUP fetch the device type slug from the API

sradevski:
I haven't tested the entire script, but I did test the snippet for setting the SLUG, and it should be fine. I'll try to get it tested tomorrow, but it can be reviewed in the meantime (which is why is marked as draft)

See: https://www.flowdock.com/app/rulemotion/resin-devices/threads/Do15pyYgb59vcSbSGVHNBD737ki
Change-type: minor
Signed-off-by: Stevche Radevski <stevche@balena.io>

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -891,7 +891,7 @@ while [[ $# -gt 0 ]]; do
             if [ -z "$2" ]; then
                 log ERROR "\"$1\" argument needs a value."
             fi
-            SLUG=$2
+            FORCED_SLUG=$2
             shift
             ;;
         --hostos-version)
@@ -963,20 +963,6 @@ fi
 
 progress 25 "Preparing OS update"
 
-if version_gt "${target_version}" "${minimum_hostapp_target_version}" || [ "${target_version}" == "${minimum_hostapp_target_version}" ]; then
-    log "Target version supports hostapps, no device type support check required."
-else
-    case $SLUG in
-        # Check board support for device types that might have 2.x-2.x non-hostapp updates
-        # The same device types listed as below in the "Switching root partition..." section
-        artik710|beaglebone*|raspberry*|intel-nuc|up-board)
-            log "Device type root partition switch is known, proceeding"
-            ;;
-        *)
-            log ERROR "Unsupported board type $SLUG."
-    esac
-fi
-
 log "Loading info from config.json"
 if [ -f /mnt/boot/config.json ]; then
     CONFIGJSON=/mnt/boot/config.json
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -997,6 +983,27 @@ UUID=$(jq -r '.uuid' $CONFIGJSON)
 API_ENDPOINT=$(jq -r '.apiEndpoint' $CONFIGJSON)
 DELTA_ENDPOINT=$(jq -r '.deltaEndpoint' $CONFIGJSON)
 
+FETCHED_SLUG=$(curl -H "Authorization: Bearer ${APIKEY}" --silent --retry 5 \
+"${API_ENDPOINT}/v6/device?\$select=is_of__device_type&\$expand=is_of__device_type(\$select=slug)&\$filter=uuid%20eq%20%27${UUID}%27" 2>/dev/null \
+| jq '.d[0].is_of__device_type[0].slug'
```
saintaardvark, line 988: This needs to be `jq -r` ; otherwise the slug is set to `"device"` (notice the quotes!) instead of `device`.

saintaardvark (commented):
Hey @sradevski -- I've tested this out on a device registered in staging.  Once I fixed the `jq` call (as noted above), I was able to upgrade everything just fine.  For reference, the sequence was:

- Install a device in staging as `test-private`
- Modify its type to `raspberrypi3`
- Try current HUP from dashboard to verify it does not work as expected; copy arguments from that invocation
- Copy your script on to device (with `jq` fix)
- Invoke your script with the same arguments
- Your script works just fine, reboot
- Verify device continues to work well
- Run HUP again from dashboard to make sure that *future* upgrades will also work
- HUP runs well, reboot
- Verify device continues to work well

This is all :+1: for me, but I'll let @xginn8 review as well.

upgrade-2.x.sh
```diff
@@ -997,6 +983,27 @@ UUID=$(jq -r '.uuid' $CONFIGJSON)
 API_ENDPOINT=$(jq -r '.apiEndpoint' $CONFIGJSON)
 DELTA_ENDPOINT=$(jq -r '.deltaEndpoint' $CONFIGJSON)
 
+FETCHED_SLUG=$(curl -H "Authorization: Bearer ${APIKEY}" --silent --retry 5 \
```
xginn8, line 986:
```suggestion
FETCHED_SLUG=$(CURL_CA_BUNDLE=${TMPCRT} curl -H "Authorization: Bearer ${APIKEY}" --silent --retry 5 \
```

xginn8, line 986:
```suggestion
FETCHED_SLUG=$(CURL_CA_BUNDLE=${TMPCRT} curl -H "Authorization: Bearer ${APIKEY}" --silent --retry 5 \
```

xginn8 (commented):
lgtm otherwise

sradevski:
@balena-ci I self-certify! since I addressed the Matt's change, I'll give this another test in staging and then we can deploy it to prod.

sradevski:
@balena-ci I self-certify!

sradevski:
@xginn8 how is balenahup "deployed"?

PR #317, opened 2020-11-12
supervisor-update: use APIKEY when querying for the new target superv…

xginn8:
…isor state

Change-type: patch


CameronDiver (approved):

xginn8:
@balena-ci rebase

PR #318, opened 2020-11-16
remove accidentally added file

xginn8:


xginn8:
@balena-ci I self-certify!

Issue #251, opened 2019-09-25
Bad API response for supervisor update breaks HUP


imrehg:
The API might give back invalid response, we should retry then, or some other way work around it still.
```
[000000940][LOG]Supervisor update: will be upgrading from v6.3.6 to v7.14.0_logstream
parse error: Invalid numeric literal at line 1, column 9
[000001968][ERROR]Couldn't extract supervisor vars...
```

But maybe this will be obsolete once the supervisor target state endpoint is used differently/not used at all? Still it's a known failure mode at the moment.

cywang117:
Closed because currently failed Supervisor upgrades do not prevent HUP from succeeding.

PR #321, opened 2021-01-04
supervisor: in case of API query failure, report the received response

xginn8:
Connects-to: https://github.com/balena-os/balenahup/issues/251
Change-type: patch


xginn8:
@balena-ci I self-certify!

PR #325, opened 2021-02-18
supervisor-upgrade: if the upgrade fails, grab what happened from sys…

xginn8:
…temd

Change-type: patch


xginn8:
@balena-ci I self-certify!

Issue #331, opened 2021-03-11
issue upgrading supervisors between 2.7.5 and 2.69.1


xginn8:
```
[000000267][LOG]Supervisor update start...
[000000267][LOG]No explicit supervisor version was provided, update to default version in target balenaOS...
Unable to find image 'registry2.balena-cloud.com/v2/765250604ca5b7831aed31728da3edf3@sha256:905b30c997a0e7b791e496e112f8312044050385b0609280dc584eacaa
3f7284' locally
docker: Error response from daemon: pull access denied for registry2.balena-cloud.com/v2/765250604ca5b7831aed31728da3edf3, repository does not exist or may require 'docker login'.
See 'docker run --help'.
```

for some reason the original hostapp image was no longer accessible, but unclear what could've caused that.

xginn8:
since older OSes get a sideloaded balenaOS hostapp, it is untagged in the local image store. when we crack that image open to find our new target supervisor, we have to contact the upstream registry and retag our local image with the registry reference. in order to do that, we must remain logged in. there is no second pull of the new OS, but just the image & layer metadata.

PR #332, opened 2021-03-11
private DTs: do not log out from the registry until cleanup

xginn8:
since older OSes get a sideloaded balenaOS hostapp, it is untagged in
the local image store. when we crack that image open to find our new
target supervisor, we have to contact the upstream registry and retag
our local image with the registry reference. in order to do that, we
must remain logged in.

Connects-to: #331
Change-type: patch


xginn8:
@balena-ci I self-certify!

Issue #338, opened 2021-04-13
If HUP fails due to sporadic network failure, progress stays at 50%


acostach:
Thus, the user is not able to determine if HUP failed or not, if it's still ongoing, etc.

Logs available in internal discussion https://www.flowdock.com/app/rulemotion/s-sales/threads/vdPYedOWjh1_Fd8aR5LYPlFzlCc
--- logs form April 2, 9, 13, issue 1).

xginn8:
this has been deployed to production

PR #339, opened 2021-04-13
logging: retry reporting failures every minute for an hour

xginn8:
Connects-to: https://github.com/balena-os/balenahup/issues/338
Change-type: patch


xginn8:
@balena-ci I self-certify!

Issue #352, opened 2021-07-29
If supervisor fails to update, hup does not resume


klutchell:
I suspect the issue is how we are printing the supervisor update logs after a failure.

https://github.com/balena-os/balenahup/blob/2824446adf45db9b68ee61cea0ada9379ff80214/upgrade-2.x.sh#L236-L237

We probably need to run this with `--no-pager`, and maybe use journalctl while we are at it.

PR #353, opened 2021-08-16
upgrade-2.x.sh: Prevent supervisor update from halting HUP

klutchell:
Rather than attempting to print logs after a failed
supervisor update just default to piping the logs during
the systemd-run command and print the warning afterwards
if required.

Resolves: https://github.com/balena-os/balenahup/issues/352
Change-type: patch


klutchell:
@balena-ci I self-certify!

alexgg (approved):
lgtm

Issue #355, opened 2021-09-10
supervisor not upgraded when updating jetson-tx2 from 2.31.2+rev1 to 2.45.1+rev3


acostach:
This is reproducible every time time, and if a subsequent hostOS update is triggered to 2.73.1, the old supervisor 9.11.1 will reboot the device durning HUP, before it is completed.

```
New hooks ran successfully.
[000000050][LOG]Running post-update fixes for jetson-tx2
[000000050][LOG]Applying extlinux.conf fix for jetson-tx2
[000000050][LOG]extlinux difference detected
[000000050][LOG]adding previous `isolcpu` value to extlinux.conf
[000000050][LOG]appending previous `APPEND` value in /mnt/boot/extra_uEnv.txt
[000000050][LOG]Supervisor update start...
[000000050][LOG]No explicit supervisor version was provided, update to default version in target balenaOS...
[000000051][LOG]Extracted default version is v10.3.7...
[000000052][LOG]Supervisor update: will be upgrading from v9.11.1 to v10.3.7
[000000053][LOG]Extracted supervisor vars: ID: 1424178
[000000053][LOG]Setting supervisor version in the API...
[000000054][LOG]Stopping supervisor and related services...
[000000058][LOG]Running supervisor updater...
systemd-run: unrecognized option '--pipe'
[000000058][WARNING]Supervisor couldn't be updated, continuing anyways
[000000059][LOG]No update package cleanup done
[000000059][LOG]Rebooting into new OS in 5 seconds...
SSH session disconnected
SSH reconnecting...
SSH reconnecting...
SSH reconnecting...
Spawning shell...



root@52a263c:~# cat /etc/os-release 
ID="balena-os"
NAME="balenaOS"
VERSION="2.45.1+rev3"
VERSION_ID="2.45.1+rev3"
PRETTY_NAME="balenaOS 2.45.1+rev3"
MACHINE="jetson-tx2"
VARIANT="Production"
VARIANT_ID="prod"
META_BALENA_VERSION="2.45.1"
RESIN_BOARD_REV="13bd883"
META_RESIN_REV="0c90c7e"
SLUG="jetson-tx2"
```

PR #356, opened 2021-09-10
upgrade-2.x.sh: Remove unsupported flag from supervisor upgrade

klutchell:
The --pipe flag is not supported in older versions of systemd like
those running on balenaOS 2.31.2+rev1.

Restore the old behaviour of runnning a separate systemctl status
command after the upgrade transient service has completed or failed.

Change-type: patch
Reverts: https://github.com/balena-os/balenahup/pull/353
Resolves: https://github.com/balena-os/balenahup/issues/355


alexgg (approved):
lgtm

Issue #354, opened 2021-09-10
balenahup doesn't apply jetson-tx2 fixes when updating from 2.31.2+rev1 to 2.73.1+rev1, SV still runs during HUP and reboots the board


acostach:
As visible in the logs in this thread, HUP is stopping the supervisor to prevent reboots triggered by the SV during the hostOS update, when updating from 2.31.2 to 2.45:
https://www.flowdock.com/app/rulemotion/r-supervisor/threads/V3mA7C4qWcm13SotzZMnxqeJSCY

```
================upgrade-2.x.sh HEADER START====================
Fri Apr  2 13:53:38 UTC 2021
[000000000][LOG]Normalized target version: 2.45.1+rev3
[000000000][LOG]Loading info from config.json
[000000001][LOG]Target version supports hostapps, no device type support check required.
[000000001][LOG]Target OS version "2.45.1+rev3" OK.
[000000001][LOG]OS variant: 2.31.2
[000000001][LOG]Host OS version "2.31.2+rev1" OK.
[000000002][LOG]Attempting host OS update using deltas
[000000006][LOG]Found delta image: registry2.balena-cloud.com/v2/2e2f9ab8fc6567991be081440e5ad047:delta-6c66dcddc15bfe37, size: 185 MB
[000000006][LOG]No resin-device-progress fix is required...
[000000006][LOG]No supervisor updater fix is required...
[000000006][LOG]hostapp-update command exists, use that for update
[000000017][WARNING]logging into registry failed, proceeding anyway (only required for private device types)
[000000017][LOG]Running pre-update fixes for jetson-tx2
[000000017][LOG]Caching current extlinux.conf for jetson-tx2 fix
[000000017][LOG]Stopping supervisor to prevent reboots during extlinux.conf updating <--
[000000017][LOG]Stopping supervisor and related services... <--
[000000028][LOG]Starting hostapp-update
```

And this is still applied when updating from 2.31.2 to to 2.45:

```
root@52a263c:/mnt/data# tail -f balenahup/upgrade-2.x.sh.20210910_123126.log 
[000000002][LOG]Host OS version "2.31.2+rev1" OK.
[000000003][LOG]Attempting host OS update using deltas
[000000006][LOG]Found delta image: registry2.balena-cloud.com/v2/2e2f9ab8fc6567991be081440e5ad047:delta-6c66dcddc15bfe37, size: 185 MB
[000000006][LOG]No resin-device-progress fix is required...
[...]

PR #357, opened 2021-09-13
upgrade-2.x: apply jetson fix to OS < v2.84.7

jakogut:
A mistyped path in the hostapp-update hooks in meta-balena caused the
extlinux.conf file to not be persisted across HUP, causing variables set
in the config to be reset by the supervisor, triggering untimely reboots
during upgrades.

A fix has been made to meta-balena [0] to properly persist the
configuration across a HUP, but this will require an OS upgrade to take
effect.

The supervisor has a similar fix in place to prevent reboots when
healthcheck breadcrumbs are present, but this requires a new supervisor
release.

To cover devices that are running older supervisor releases on older OS
versions, extend the version check to apply the Jetson fix to affected
OS versions.

[0] https://github.com/balena-os/meta-balena/commit/096024808224aea2100e2a2a1f95ef75e23493ef

Change-type: patch
Signed-off-by: Joseph Kogut <joseph@balena.io>

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -527,7 +527,7 @@ function hostapp_based_update {
             ;;
         jetson-tx2)
             log "Running pre-update fixes for ${SLUG}"
-            if version_gt "${HOST_OS_VERSION}" "2.31.1" && version_gt "2.58.3" "${target_version}" ; then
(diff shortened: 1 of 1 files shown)


acostach (approved):
Looks good to me.

Related discussion https://github.com/balena-os/balenahup/issues/354#issuecomment-917958848

PR #360, opened 2021-09-17
Temporarily disable upgrades to 2.83.10+rev1 for Pi4

klutchell:
Change-type: patch
See: https://jel.ly.fish/51fc3ffe-33de-40c4-8982-abeb89717d08


klutchell:
@balena-ci I self-certify!

PR #362, opened 2021-10-04
Backport increased os_bc_limit to existing releases

acostach:
Starting with BalenaOS v2.85.9 the u-boot bootcount limit
has been increased to 3. Let's apply this fix previous
jetson-tx2 releases, like 2.73 or 2.82.

Change-type: patch
Signed-off-by: Alexandru Costache <alexandru@balena.io>

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -363,6 +363,13 @@ function post_update_jetson_fix {
             echo "extra_os_cmdline=${APPEND_value}" >> "${uEnv_file}" && sync "${uEnv_file}"
         fi
     fi
+
+    if [ -e "${uEnv_file}" ] && grep -q '^os_bc_lim=' "${uEnv_file}"; then
+        log "Fix for jetson-tx2 bootcount limit already applied"
+    else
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -363,6 +363,13 @@ function post_update_jetson_fix {
             echo "extra_os_cmdline=${APPEND_value}" >> "${uEnv_file}" && sync "${uEnv_file}"
         fi
     fi
+
+    if [ -e "${uEnv_file}" ] && grep -q os_bc_lim "${uEnv_file}"; then
```
klutchell, line 367:
```suggestion
    if [ -e "${uEnv_file}" ] && grep -q '^os_bc_lim=' "${uEnv_file}"; then
```
I haven't tested this, but I think we can be more specific with our grep here.

acostach: Sure, done @klutchell

klutchell (approved):
lgtm

acostach:
Thank you @klutchell , please let me know when it's available for testing in staging

PR #364, opened 2021-10-26
upgrade-2.x: Allow to upgrade OS with custom image

quentin-burg:
Allow to update an OS with a custom image located on a DockerHub private registry by providing parameters starting by `--private`


pmirabel:
Is there any chances that this feature will be added soon ?

PR #374, opened 2022-09-19
upgrade-2.x: Remove REC files before updating

jaomaloy:
REC files are generated by fsck and fills up
the boot partition quickly which causes failures
in updates. Clean this up before proceeding with
the update.

Change-type: patch
Signed-off-by: jaomaloy <jao.maloy@balena.io>

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -170,6 +170,16 @@ function remove_containers() {
     ${DOCKER_CMD} rm $(${DOCKER_CMD} ps -a -q) > /dev/null 2>&1 || true
 }
 
+function remove_rec_files() {
+    local boot_dir='/mnt/boot'
+    shopt -s nullglob
+    for f in ${boot_dir}/*.REC; do
+        log WARN "Removing $f from boot partition"
+        rm -f $f
+    done
+    sync ${boot_dir}
+}
+
 #######################################
 # Upgrade the supervisor on the device.
 # Extract the supervisor version with which the the target hostOS is shipped,
@@ -526,6 +536,9 @@ function hostapp_based_update {
         storage_driver=$(cat /boot/storage-driver)
     fi
 
+    # remove REC files on boot partition
+    remove_rec_files
+
     case ${SLUG} in
         raspberry*)
             log "Running pre-update fixes for ${SLUG}"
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -170,6 +170,16 @@ function remove_containers() {
     ${DOCKER_CMD} rm $(${DOCKER_CMD} ps -a -q) > /dev/null 2>&1 || true
 }
 
+function remove_rec_files() {
+    echo "Removing REC files to make space for the update."
+    local boot_dir='/mnt/boot'
+    for f in ${boot_dir}/*.REC; do
+        log "Removing $f from boot partition"
+        rm -f $f
+    done
```
alexgg, line 179: The code above will print a message even if there are no REC files:
For example, 
```
tmpdir=$(mktemp -d)
for f in ${tmpdir}/*.REC; do echo $f; done
/tmp/tmp.RUcM4kNJiB/*.REC
```

We  want to add a warning message only if REC files are present as they should not be, so maybe something like the following works better:

```
find "${boot_dir}" -name "*.REC" -exec log WARN "Removing {} from boot partition" \; -exec  rm -f {} \;
```

alexgg (changes requested):

upgrade-2.x.sh
```diff
@@ -170,6 +170,16 @@ function remove_containers() {
     ${DOCKER_CMD} rm $(${DOCKER_CMD} ps -a -q) > /dev/null 2>&1 || true
 }
 
+function remove_rec_files() {
+    echo "Removing REC files to make space for the update."
```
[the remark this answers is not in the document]
jaomaloy: removed.

upgrade-2.x.sh
jaomaloy: `-exec log WARN "Removing {} from boot partition" \;` has trouble reading the file. Tested a few more variants before settling on using `shopt -s nullglob` with the original loop. Let me know if there are better options

alexgg (approved):
lgtm

alexgg:
@balena-ci retest

alexgg:
@jaomaloy I amended the commit to allow versionbot to run without errors.

PR #384, opened 2023-09-27
Fix race condition with supervisor self-serve updates

alexgg:


alexgg:
Test plan:

- [ ] Golden path - SV is updated to a newer version
- [ ] Race: Self-serve SV update changes version and bad request for downgrade is received
- [ ] Error: A request to update to an unknown SV version is made and a bad request is received

upgrade-2.x.sh
```diff
@@ -215,6 +215,19 @@ function _run_supervisor_update() {
     return "${ret}"
 }
 
+function _fetch_supervisor_version() {
+    local resp
+    local supervisor_version
+
+    resp=$(CURL_CA_BUNDLE=${TMPCRT} curl --silent --retry 10 --header "Authorization: Bearer ${APIKEY}" "${API_ENDPOINT}/v6/device(uuid='${UUID}')?\$select=supervisor_version")
```
thgreasi, line 222: Heads-up that the check in the API compares the current `device.should_be_managed_by__supervisor_release` (not the `device.supervisor_version`, even though someone could argue we should test both) and compares it with the new targer supervisor release.
See: https://github.com/balena-io/open-balena-api/blob/d632e075e543bd7c8125fe97559631efbcb56831/src/features/supervisor-app/hooks/supervisor-app.ts#L101-L131

alexgg: thanks @thgreasi I pushed a new version that checks `should_be_managed_by__supervisor_release` and aborts if its newer than the `supervisor_version` as we don't want two simultaneous updates going on.

alexgg:
About inheriting ERR traps by subshells, we have seen this error on support:
```
++ DOCKER_HOST=unix:///var/run/balena-host.sock                                                                       
++ balena run --rm registry2.balena-cloud.com/v2/e2c8d851f80e8074d202306ddad84f0d@sha256:4700800c9477d18d3cba67f6a75c3
0c4bf7453cc9dafd3719d7f2f395e4e1b77 bash -c 'cat /etc/*-supervisor/supervisor.conf | sed -rn '\''s/SUPERVISOR_(TAG|VER
SION)=v(.*)/\2/p'\'''                                                                                                 
docker: Cannot connect to the balenaEngine daemon at unix:///var/run/balena-host.sock. Is the balenaEngine daemon runn
ing?.                                                                                                                 
See 'docker run --help'.                                                                                              
+ DEFAULT_SUPERVISOR_VERSION=                                                                                         
+ _no_more_locking                                                                                                    
+ _lock u                                                                                                             
+ flock -u 99                                                                                                         
+ _lock xn                                                                                                            
+ flock -xn 99                                                                                                        
+ rm -f /var/lock/resinhup.lock                                                                                       
+ rm -f /tmp/tmp.N9KzNGuJAu                                                                                           
+ rm -f /tmp/tmp.M2AjOgLELW                                                                                           
root@f2fc008:/tmp# echo $?                                                                                            
125                     
```

Which silently fails leaving the device in updating state.

With this modification, the subshell error correctly sets the failed state in the cloud:
```++ DOCKER_HOST=unix:///var/run/balena-host.sock                                                                                                                                                                                             
++ balena run --rm registry2.balena-cloud.com/v2/e2c8d851f80e8074d202306ddad84f0d@sha256:4700800c9477d18d3cba67f6a75c3                                                                                                                      
0c4bf7453cc9dafd3719d7f2f395e4e1b77 bash -c 'cat /etc/*-supervisor/supervisor.conf | sed -rn '\''s/SUPERVISOR_(TAG|VER                                                                                                                      
SION)=v(.*)/\2/p'\'''                                                                                                                                                                                                                       
docker: Cannot connect to the balenaEngine daemon at unix:///var/run/balena-host.sock. Is the balenaEngine daemon runn                                                                                                                      
ing?.                                                                                                                                                                                                                                       
See 'docker run --help'.                                                                                                                                                                                                                    
+++ report_update_failed                                                                                                                                                                                                                    
+++ perc=100                                                                                                                                                                                                                                
+++ state='OS update failed'                                                                                                                                                                                                                
+++ compare_device_state 100 'OS update failed'                                                                                                                                                                                             
+++ perc=100                                                                                                                                                                                                                                
+++ state='OS update failed'                                                                                                                                                                                                                
+++ local resp                                                                                                                                                                                                                              
+++ local remote_perc   
...

```

upgrade-2.x.sh
```diff
@@ -215,6 +215,34 @@ function _run_supervisor_update() {
...
+    local scheduled_supervisor_version
+
+    resp=$(CURL_CA_BUNDLE=${TMPCRT} curl --silent --retry 10 --header "Authorization: Bearer ${APIKEY}" "${API_ENDPOINT}/v6/device(uuid='${UUID}')?\$select=supervisor_version")
+    if supervisor_version=$(echo "${resp}" | jq -e -r '.d[0].supervisor_version'); then
+        if [ -z "${supervisor_version}" ]; then
+            log ERROR "Could not get current supervisor version from the API, got ${resp}"
+            return 1
+        fi
+        # Make sure there is no scheduled supervisor update
+        resp=$(CURL_CA_BUNDLE=${TMPCRT} curl --silent --retry 10 --header "Authorization: Bearer ${APIKEY}" "${API_ENDPOINT}/v6/device(uuid='${UUID}')?\$select=should_be_managed_by__supervisor_release&\$expand=should_be_managed_by__supervisor_release(\$select=supervisor_version)")
```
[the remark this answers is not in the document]
thgreasi: @alexgg a) you could get just the supervisor_release's supervisor_version with:
```
curl "${API_ENDPOINT}/v6/supervisor_release?\$top=1&\$select=supervisor_version&\$filter=should_manage__device/any(d:d/uuid%20eq%20'${UUID}')"
```
but I think it would be better if you tried to combine both curls and fetch both the `device.supervisor_version` and the supervisor_release's supervisor_version in one request with
```suggestion
        resp=$(CURL_CA_BUNDLE=${TMPCRT} curl --silent --retry 10 --header "Authorization: Bearer ${APIKEY}" "${API_ENDPOINT}/v6/device(uuid='${UUID}')?\$select=should_be_managed_by__supervisor_release&\$expand=should_be_managed_by__supervisor_release(\$select=supervisor_version)")
        resp=$(CURL_CA_BUNDLE=${TMPCRT} curl --silent --retry 10 --header "Authorization: Bearer ${APIKEY}" "${API_ENDPOINT}/v6/device(uuid='${UUID}')?\$select=supervisor_version&\$expand=should_be_managed_by__supervisor_release(\$top=1;\$select=supervisor_version)")
```

That's both 1 less request & and is also faster then the query that starts from the `supervisor_release` (I expect that part of the reason is that the `supervisor_release` is a translation and also has some expensive permissions).

b) The API does support compressed request/responses, so given that it saves bandwidth I would use it.
The supervisor is doing so in all requests already for example and last year was even migrated to use brotli.

alexgg:
Re-test:

- [ ]  Golden path - SV is updated to a newer version without self-serve update
- [ ]  Golden path - SV is updated to a newer version after self-serve update
- [ ]  HostOS update success, but SV is not updated as it already has a newer version
- [ ]  Race: Self-serve SV update changes version and bad request for downgrade is received

thgreasi (commented):
:+1: :heart:

alexgg:
lgtm

PR #385, opened 2023-10-12
upgrade-2.x: perform a safe reboot taking locks into account

alexgg:
When a reboot is requested take into account that user applications might be in critical sections and might have requested for the reboot to wait.

Change-type: patch


alexgg:
lgtm

PR #387, opened 2023-11-29
upgrade-2.x: do not fail update if there is a scheduled SV update

alexgg:
Currently, HUP fails if there is a scheduled supervisor update pending in the cloud to avoid concurrent updates.

This has caused a race condition as follows:

* Users start both a HUP and a SV update as the buttons are close to each other and it's not that clear HUP will update the SV too
* As soon as the SV update is trigerred, the target state is patched by the actions server. target state.
* However, the supervisor upgrade is aborted in the actions server as it checks whether the device is idle (and it is not because of HUP): ``` actions-backend[8486]: 331a727bc378acf3cd86f305b917bdc16eada2c468fe3a0c41df5ab475be21fe POST actions.balena-devices.com /v1/8412784dc5bf1d5f1a78c5f6d6fbb5e1/supervisorupgrade 500 52 25.528ms ```
* HUP fails
* HUP cannot be retried until the SV updates - which now only happens in a timer once a day or on reboot

This change modifies the supervisor update logic to take into account the provided target version, and the current and scheduled target state version and acts accordingly.

Change-type: patch

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -233,6 +233,18 @@ function _run_supervisor_update() {
     return "${ret}"
 }
 
+# Fetch the current and scheduled supervisor versions from the API
+#
+# Returns:
+#
+#  0: Success
+#  1: Failure
+#
+#  Outputs:
+#
+#  On success, a string separated string of current and scheduled supervisor
+#  versions.
+#
 function _fetch_supervisor_version() {
     local resp
     local supervisor_version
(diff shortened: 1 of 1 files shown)


alexgg:
Manual tests:

- [ ] Update to new target supervisor version - golden path
  * Target state version is updated to target
  * Supervisor is updated to target
- [ ] Update to target version newer than current but not newer than scheduled
  * No patch as scheduled version is already newer
  * Supervisor is updated to scheduled
- [ ] Update to target version newer than scheduled but not newer than current
  * In target state, current is newer than scheduled - effectively a SV downgrade that should not happen
  * Patch target version to current so it does not downgrade
  * No need to update
- [ ] Update to target version not newer than current nor scheduled
 * No need to update
- [ ]

upgrade-2.x.sh
```diff
@@ -352,26 +363,44 @@ function upgrade_supervisor() {
...
-                # No need to stop services as supervisor update will do that for us
-                if _run_supervisor_update; then
-                    if version_gt "6.5.9" "${target_supervisor_version}" ; then
-                        remove_containers
-                        log "Removing supervisor database for migration"
-                        rm /resin-data/resin-supervisor/database.sqlite || true
-                    fi
-                else
-                    log WARN "Failed to update supervisor to target state version - leave to first boot."
+    if supervisor_target_state_versions=$(_fetch_supervisor_version); then
```
klutchell, line 366: is this really a valid if condition in bash? It's not actually checking that they are equal, just that the subshell did not fail, correct?

alexgg: That's right @klutchell it's checking the return code of the _fetch_supervisor_version.

PR #389, opened 2024-03-28
Upgrade-2.x.sh: Comment why we still need to use the release_tags

thgreasi:
Change-type: patch
See: https://balena.fibery.io/Work/Project/Use-the-release-semver-fields-along-with-release_tags-across-all-our-components-404


klutchell:
LGTM

PR #391, opened 2024-04-25
Attempt to report unhandled errors as failures to cloud

alexgg:


klutchell:
LGTM

alexgg:
Manual tests against BM:

Problem reproduction:

* Trigger update and disconnect while image is being downloaded
* wait a couple of minutes
* Reconnect - allow engine to resume download
* Engine finishes download but reports error
* The update script hangs without exiting and reporting failure or removing locks

After the patch, the script correctly ends and reports failure.

Also tested:

- [ ] If BM is not reachable to report failure, the script waits and keeps trying using curl's default backoff algorithm
- [ ] Manually launching the script and interrupting it at different points correctly reports failure to cloud

PR #407, opened 2024-11-20
Add takeover based HUP script

kb2ma:
This script is equivalent to `upgrade-2.x.sh`, but for takeover based HUP. This script is derived from `upgrade-2.x.sh`, but removes unnecessary functionality. It also aims to be a model for a script implementation but avoiding gratuitous changes to the `upgrade-2.x.sh` original.

I am looking for higher level comments on the approach at this point. I have added some questions in the code. Don't mind the individual commits. I will squash them later.

### DIFF
--- a/upgrade-takeover.sh
+++ b/upgrade-takeover.sh
@@ -1,0 +1,494 @@
+#!/bin/bash
+#
+# Upgrades balenaOS version by running takeover to flash storage, potentially
+# including repartitioning. See https://github.com/balena-os/takeover
+#
+# Arguments:
+#  * --hostos-version is required to specify target OS version
+#  * See help() below for details
+#
+# Outputs:
+#  * Writes log file to /mnt/boot/balenahup
+#  * Notifies balenaAPI by PATCHing the device via /usr/bin/resin-device-progress script
+#
+# Results:
+#  * Flashes new OS and reboots into it on success
+#  * Returns 9 if another upgrade script may already be running
+#  * Returns 1 on other errors in this script
+#  * Reboots into current OS if fails later in the attempt
+#
+# Notes:
+# This script was derived from upgrade-2.x.sh for traditional balenahup upgrades.
+
+SCRIPTNAME=upgrade-takeover.sh
+
+# Set up cautious error handling, run ERR trap on failure
+set -o errexit
+set -E
+set -o pipefail
+
+# Define variables about running balenaOS, including VERSION and SLUG (device type)
+# shellcheck disable=SC1091
+source /etc/os-release
+
+# Set PATH for binary lookups
+# shellcheck disable=SC1091
+source /etc/profile
+
+# Prevent running multiple instances of upgrade
+LOCKFILE="/var/lock/resinhup.lock"
+LOCKFD=99
+# Private functions
+_lock()             { flock "-$1" $LOCKFD; }
+_exit_handler() {
+    _exit_status=$?
+    if [ "${_exit_status}" -ne 0 ]; then
+        log "Exit on error ${_exit_status}"
+        if ! report_upgrade_failed > /dev/null 2>&1; then
+            log "Failed to report progress on exit with status $?"
+        fi
+        if [ "${_exit_status}" -eq 9 ]; then
+           log "No concurrent updates allowed - lock file in place."
+        fi
+    fi
+    _no_more_locking
+    log "Lock removed - end."
+}
+_no_more_locking()  { _lock u; _lock xn && rm -f $LOCKFILE;rm -f "${outfifo}";rm -f "${errfifo}"; }
+_prepare_locking()  { eval "exec $LOCKFD>\"$LOCKFILE\""; trap _exit_handler EXIT; }
+# Public functions
+exlock_now()        { _lock xn; }  # obtain an exclusive lock immediately or fail
+
+# Helper function to report progress to device API
+# $1 -- integer: completion percentage
+# $2 -- string: detail message
+function progress {
+    pct=$1
+    message=$2
+    resin-device-progress --percentage "${pct}" --state "${message}" > /dev/null || true
+}
+
+function help {
+    cat << EOF
+Upgrade balenaOS via takeover tool
+
+Options:
+  -h, --help
+        Display this help and exit.
+
+  --force-slug <SLUG>
+        Override slug detection and force this slug to be used for the script.
+
+  --hostos-version <HOSTOS_VERSION>
+        Run the updater for this specific HostOS version as semver or ESR, where
+        semver is in the format major.minor.patch, like 2.5.1; and ESR is in the
+        format year.month.patch, like 2024.4.0. The version must begin with a
+        digit, not a 'v'. This is a mandatory argument.
+
+  --balenaos-registry
+        Unused but accepted for compatibility with upgrade-2.x.sh when called
+        by balena-proxy.
+EOF
+}
+
+# Notify backend device API that upgrade has failed. This report is essential
+# to allow a user to retry the upgrade. Sends the report, and then independently
+# verifies success. Tries once per minute for an hour.
+function report_upgrade_failed() {
+    pct=100
+    state="OS update failed"
+    while ! compare_device_state "${pct}" "${state}"; do
+        ((c++)) && ((c==60)) && break
+        if resin-device-progress --percentage "${pct}" --state "${state}"; then
+            continue
+        fi
+        log WARN "Retrying failure report - try $c"
+        sleep 60
+    done
+}
+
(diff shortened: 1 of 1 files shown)


upgrade-takeover.sh
```diff
@@ -0,0 +1,460 @@
...
+#   $target_version; already verified
+#   $SLUG; already verified
+#   $takeover_page; already verified
+# Exits script on download failure
+function download_target_image() {
+    log "Downloading target image"
+    # Don't use --fail; we want the status code
+    curl_no_fail="curl --silent --retry 10 --location --compressed"
+
+    status_code=$(\
```
kb2ma, line 234: Worthwhile to capture status code?

upgrade-takeover.sh
```diff
@@ -0,0 +1,460 @@
...
+    *)
+        log ERROR "Host OS version \"$VERSION\" not supported."
+        ;;
+esac
+
+# Ensure takeover environment is prepared
+takeover_path="/mnt/data/takeover"
+mkdir -p "${takeover_path}"
+
+# Download target hostOS image as required
```
kb2ma, line 433: Is it feasible to avoid re-downloading the same OS image in case of running again after failure?

upgrade-takeover.sh
```diff
@@ -0,0 +1,460 @@
...
+    [2-9].*|[1-9][0-9].*|2[0-9][0-9][0-9].*.*)
+        log "Host OS version \"$VERSION\" OK."
+        ;;
+    *)
+        log ERROR "Host OS version \"$VERSION\" not supported."
+        ;;
+esac
+
+# Ensure takeover environment is prepared
+takeover_path="/mnt/data/takeover"
```
kb2ma, line 430: Is it reasonable to copy the takeover binary and OS image to this directory. OS image is approximately 250 MB.

upgrade-takeover.sh
```diff
@@ -0,0 +1,460 @@
...
+mkdir -p "${takeover_path}"
+
+# Download target hostOS image as required
+if [ ! -f "${takeover_path}/balenaos.img.gz" ]; then
```
alexgg, line 434: How do you know this is the same as the requested version? I would honestly just cleanup on exit and re-download.

upgrade-takeover.sh
```diff
@@ -0,0 +1,460 @@
...
+    download_target_image
+else
+    log "balenaOS target image available; download not required"
+fi
+
+# Download takeover binary as required
+if [ ! -f "${takeover_path}/takeover" ]; then
+    download_takeover_binary "v0.9.0-dev.1"
```
alexgg, line 442: Why not always download the latest version? Would we ever release a new version that we don't want to use?

upgrade-takeover.sh
```diff
@@ -0,0 +1,460 @@
...
+else
+    log "Takeover binary available; download not required"
```
alexgg, line 444: I would clean-up on exit and always re-download.

upgrade-takeover.sh
```diff
@@ -0,0 +1,460 @@
...
+# Run takeover
+# Must run from a writable directory; takeover creates temp files there
+cd ${takeover_path}
+
+# No need to specify config.json path; defaults to /mnt/boot.
+# API check fails on BoB, so disabled
+# We do *not* expect takeover to return on success; it flashes the device with
+# the new balenaOS.
+res=$(./takeover -i balenaos.img.gz \
+   --no-ack --no-nwmgr-check --no-os-check --no-vpn-check \
```
alexgg, line 458: Why no-vpn-check? Isn't it better to default to only update if the device has a fair chance of being remotely accessible?

upgrade-takeover.sh
```diff
@@ -0,0 +1,460 @@
...
+   --log-level debug -l /dev/sda1 --s2-log-level debug)
```
alexgg, line 459: Is `-l /dev/sda1`  the target device? If so it should not be hard coded as different devices will have different storage. It should be deduced from the running system I reckon, like inspecting the host OS to see what parent device is the active partition mounted on.

alexgg (changes requested):

upgrade-takeover.sh
alexgg: Probably useful in the log to know the cause for failure.

upgrade-takeover.sh
```diff
@@ -0,0 +1,460 @@
...
+
+    ${CURL} -o "${takeover_path}/takeover.tar.gz" "$download_url" || log ERROR "Could not download takeover binary, aborting."
+
+    # Extract and prepare for use
+    tar -C "${takeover_path}" -zxvf "${takeover_path}/takeover.tar.gz"
+    chown root:root "${takeover_path}/takeover"
+    rm "${takeover_path}/takeover.tar.gz"
+}
+
+# Download target balenaOS image; expects a public image
```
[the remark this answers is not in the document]
kb2ma: I understand the API download endpoint _does_ support private images.  I am confirming with those in the know since I don't have such a device. I'll update the comment.

upgrade-takeover.sh
kb2ma: I see that I can get the content-length header from the API call. Is the best approach to compare against the output from `df` for the partition mounted on `/mnt/data` -- or is there an easier way to grab that available size?

upgrade-takeover.sh
```diff
@@ -0,0 +1,460 @@
...
+    log "Stopping supervisor and related services..."
+    systemctl stop update-balena-supervisor.timer > /dev/null 2>&1 || systemctl stop update-resin-supervisor.timer > /dev/null 2>&1
+    systemctl stop balena-supervisor  > /dev/null 2>&1 || systemctl stop resin-supervisor > /dev/null 2>&1
+    ${DOCKER_CMD} rm -f balena_supervisor resin_supervisor > /dev/null 2>&1 || true
+}
+
+# Remove and log fsck recovery files. Since takeover will reflash the disk, it's
+# not strictly necessary. However, it may indicate issues with the storage medium,
+# so good to know.
+function remove_rec_files() {
```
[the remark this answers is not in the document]
kb2ma: How does `fsck` help in this case if takeover will overwrite all of the partitions?

upgrade-takeover.sh
kb2ma: I agree. This version is a pre-release just for testing until we have a binary that is reasonably useful, so it can be called 'latest'.

upgrade-takeover.sh
kb2ma: @rahul-thakoor , any concerns with including the VPN check? Should we set `--check-timeout`? What is your experience?

kb2ma: @acostach ^^

upgrade-takeover.sh
kb2ma: Yes, that is for logging to an external device. That was just there until we incorporated the on-device logging. See balena-os/takeover#100

upgrade-takeover.sh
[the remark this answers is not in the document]
alexgg: I was following up on the comment above - `fsck` or similar would show whether the media is healthy before copying and launching the takeover tool. How takeover would fail from a damaged storage is difficult to predict - probably failing to write and handling the error, but it's probably an untested path.

However, if it fails to write because of damaged sectors, it could have already overwritten parts so the device will be bricked. Bailing out if fsck errors is a way to improve reliability.

upgrade-takeover.sh
alexgg: Something like `df --output=avail /mnt/data/ | tail -n 1` should work well enough

upgrade-takeover.sh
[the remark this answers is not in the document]
kb2ma: Sounds fine to me. So to be clear, you're saying use `fsck -n`, capture stdout+stderr, and failure is defined as any non-zero result. Log the output on failure. Use `lsblk` to find the device name of the `/mnt/data` partition?

Let me know of any better ways; not my expertise.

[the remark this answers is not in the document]
alexgg: Yeah, something like:
```
fsck.ext4 -n /dev/sda1 > dev/null
if [ $? -ge 4 ]; then
    exit 1
fi
```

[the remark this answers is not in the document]
kb2ma: Done; to be clear, only checking data partition.

kb2ma:
@alexgg Fixed up from your review and rebased. All changes are in the final commit. In addition to your review, I did these things:

* Renamed "takeover_path" to "work_dir". It's more descriptive, especially since adding the cleanup function
* Added some options to the takeover binary to accommodate how it works now

alexgg (approved):

PR #424, opened 2026-03-24
Require target OS version is an upgrade

kb2ma:
The SDK has added the ability to pin to an OS version, which means the actual update happens asynchronously on the device. In addition, the user may pin back to the current OS version.

These changes mean a race condition on the pinned version may occur between the cloud and the device. This PR ensures the device by default requires the target version is an upgrade.

However, also added a `--no-require-upgrade` option to avoid this requirement. We expect its use only for manual updates or support use. Remember that the automation around the pinned OS version means that a downgrade on the device will be reversed by the automation unless the API pin also is lowered.

Change-type: minor

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -1041,9 +1041,19 @@ if [ -n "$target_version" ]; then
         if ! version_gt "$target_version" "$minimum_target_version" &&
             ! [ "$target_version" == "$minimum_target_version" ]; then
                 log ERROR "Target OS version \"$target_version\" too low, please use \"$minimum_target_version\" or above."
-                else
+            else
+                # Strip the pre-release portion of the target raw_version, based on
+                # the format "<major>.<minor>.<patch>[-<pre-release>][+<revision>]".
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -143,6 +147,20 @@ function log {
...
+function version_strip_prerelease() {
+    vpre=$(echo "$1" |cut -d "-" -f 1)
+    # If suffix missing, print ""
+    vsuf=$(echo "$1" |cut -s -d "+" -f 2)
+
+    if [ "$vsuf" = "" ]; then
+        echo "${vpre}"
+    else
+        echo "${vpre}+${vsuf}"
+    fi
```
thgreasi: As it is atm it duplicates the rev part (eg 6.10.22+rev1 => 6.10.22+rev1+rev1).
Since we are already using sed in other places of the HUP script, how about using sed w/ a simple regex (which is easier for me understand). eg:
```suggestion
    echo "$1" | sed -E 's/-[^+]+(\+|$)/\1/'
```

thgreasi (approved):
:sparkles:

PR #427, opened 2026-04-08
Support upgrade requirement from rolling to ESR releases

kb2ma:
Must verify that an update from a rolling version (like 6.5.4) to an ESR version (like 2025.7.0) is an upgrade -- a more recent release. To compare ESR and rolling releases, must determine the rolling version corresponding to a provided ESR version. This determination is via the release API. Also refactored the API query since we already query the API for image location.

Also includes a commit to remove some code duplication around the target version.

Part of project https://balena.fibery.io/Work/Project/Helios-MLP-1704

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -574,7 +574,7 @@ function hostapp_based_update {
             ;;
         jetson-tx2)
             log "Running pre-update fixes for ${SLUG}"
-            if version_gt "${HOST_OS_VERSION}" "2.31.1" && version_gt "2.84.7" "${target_version}" ; then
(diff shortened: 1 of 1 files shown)


thgreasi:
Can we plz change the commit and PR title to something like v so that it's clear that we only support rolling -> ESR
> Support upgrade requirement from rolling to ESR releases

or
> Fix the upgrade requirement check for rolling to ESR OS updates

The current "between ESR and rolling releases" sounds (to me) like we support HUPs between them on both directions.

thgreasi:
How about v ?
Block rolling to ESR OS updated that would be a meta-balena-base downgrade

PR #428, opened 2026-04-21
Disallow target with balenaEngine v25 if using aufs storage

kb2ma:
First must upgrade to an earlier version to convert to overlayfs.

Change-type: patch

Determination of balenaEngine version based on `balena info` output:
```
root@14a942a:~# balena info
Containers: 1
 Running: 1
 Paused: 0
 Stopped: 0
Images: 1
Server Version: 17.12.0-dev
Storage Driver: aufs
...
```

### DIFF
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -14,7 +14,12 @@ set -o pipefail
 minimum_hostos_version=2.14.0
 minimum_target_version=2.16.0
 minimum_supervisor_stop=2.53.10
-minimum_aufs_unsupported=8.0.0
+# Metabalena OS version at which aufs support was removed from the included balenaEngine,
+# and minimum version that provides migration to overlayfs.
+# Provides an interval of versions to which such a device must hop before updating
+# to latest release.
+aufs_removed_version=8.0.0
+aufs_minimum_migration=2.85
 
 # This will set VERSION, SLUG
 # shellcheck disable=SC1091
@@ -1045,28 +1050,40 @@ fi
 
 if [ -n "$target_version" ]; then
     case $target_version in
-        [2-9].*|2[0-9][0-9][0-9].*.*)
-        if ! version_gt "$target_version" "$minimum_target_version" &&
-            ! [ "$target_version" == "$minimum_target_version" ]; then
+        2[0-9][0-9][0-9].*.*)
+            # Metabalena ESR OS version at which aufs support was removed from
+            # the included balenaEngine, and minimum version that provides migration
+            # to overlayfs.
+            # Overwrites the variable values defined at the top of the script,
+            # which are based on rolling OS version.
+            aufs_removed_version=2026.7.0
+            aufs_minimum_migration=2022.1
+            ;;
+        [2-9].*)
+            # Note: Update comparison to include ESR target when minimum target
+            # version advances to at least v2.44, when ESR introduced.
+            if ! version_gt "$target_version" "$minimum_target_version" && \
+                ! [ "$target_version" == "$minimum_target_version" ]; then
                 log ERROR "Target OS version \"$target_version\" too low, please use \"$minimum_target_version\" or above."
-            else
-                # Strip the pre-release portion of the target raw_version, based on
-                # the format "<major>.<minor>.<patch>[-<pre-release>][+<revision>]".
-                # Otherwise, version_gt would consider 1.2.3 < 1.2.3-1234. This strip
-                # also allows 1.2.3+rev1 -> 1.2.3-1234+rev2 HUPs. Although the rest
-                # of the platform treats a pre-release version as lower, ignoring
-                # the pre-release portion is the correct behavior for balenaOS versioning.
-                target_nopre=$(echo "$target_version" | sed -E 's/-[^+]+(\+|$)/\1/')
-                if [ "$REQUIRE_UPGRADE" = "yes" ] && ! version_gt "$target_nopre" "$VERSION"; then
-                    log ERROR "Target OS version \"$target_version\" must be greater than current version."
-                fi
-                log "Target OS version \"$target_version\" OK."
             fi
             ;;
         *)
             log ERROR "Target OS version \"$target_version\" not supported."
             ;;
     esac
+
+    # Verify target version is an upgrade.
+    # First strip the pre-release portion of the target raw_version, based on
+    # the format "<major>.<minor>.<patch>[-<pre-release>][+<revision>]".
+    # Otherwise, version_gt would consider 1.2.3 < 1.2.3-1234. This strip
+    # also allows 1.2.3+rev1 -> 1.2.3-1234+rev2 HUPs. Although the rest
+    # of the platform treats a pre-release version as lower, ignoring
+    # the pre-release portion is the correct behavior for balenaOS versioning.
+    target_nopre=$(echo "$target_version" | sed -E 's/-[^+]+(\+|$)/\1/')
+    if [ "$REQUIRE_UPGRADE" = "yes" ] && ! version_gt "$target_nopre" "$VERSION"; then
+        log ERROR "Target OS version \"$target_version\" must be greater than current version."
+    fi
+    log "Target OS version \"$target_version\" OK."
 else
     log ERROR "No target OS version specified."
 fi
@@ -1079,10 +1096,13 @@ if [ "${SLUG}" = "raspberrypi4-64" ] && \
     fi
 fi
 
-if version_gt "${target_version}" "${minimum_aufs_unsupported}" || \
-    [ "$target_version" == "$minimum_aufs_unsupported" ] && \
-    $(balena info |grep "Storage Driver: aufs"); then
-    log ERROR "BalenaEngine aufs storage supported only before ${minimum_aufs_unsupported}; first upgrade to at least 2.83 to convert"
+# If target OS version uses a balenaEngine that does not support aufs storage,
+# verify device is not currently using aufs. Otherwise must migrate first.
+if $(version_gt "${target_version}" "${aufs_removed_version}") ||
+      [ "$target_version" = "$aufs_removed_version" ] &&
+      $(balena info |grep "Storage Driver: aufs"); then
+    log ERROR "This device is using the AUFS storage driver and must be migrated before upgrading to v${aufs_removed_version} or later. \
+Upgrade to a release between v${aufs_minimum_migration} and < v${aufs_removed_version}."
 fi
 
 # Already retrieved if script inputs from App UUID and release commit.
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -14,12 +14,6 @@ set -o pipefail
 minimum_hostos_version=2.14.0
 minimum_target_version=2.16.0
 minimum_supervisor_stop=2.53.10
-# Metabalena OS version at which aufs support was removed from the included balenaEngine,
-# and minimum version that provides migration to overlayfs.
-# Provides an interval of versions to which such a device must hop before updating
-# to latest release.
-aufs_removed_version=8.0.0
-aufs_minimum_migration=2.85
 
 # This will set VERSION, SLUG
 # shellcheck disable=SC1091
@@ -1049,17 +1043,19 @@ if [ -n "$app_uuid" ]; then
 fi
 
 if [ -n "$target_version" ]; then
+    # Define meta-balena OS version at which aufs support was removed from the
+    # included balenaEngine, and minimum version that provides migration to overlayfs.
+    # Provides an interval of versions to which such a device must hop before
+    # updating to latest release.
     case $target_version in
         2[0-9][0-9][0-9].*.*)
-            # Metabalena ESR OS version at which aufs support was removed from
-            # the included balenaEngine, and minimum version that provides migration
-            # to overlayfs.
-            # Overwrites the variable values defined at the top of the script,
-            # which are based on rolling OS version.
             aufs_removed_version=2026.7.0
             aufs_minimum_migration=2022.1
             ;;
         [2-9].*)
+            aufs_removed_version=8.0.0
+            aufs_minimum_migration=2.85
+
             # Note: Update comparison to include ESR target when minimum target
             # version advances to at least v2.44, when ESR introduced.
             if ! version_gt "$target_version" "$minimum_target_version" && \
@@ -1100,7 +1096,7 @@ fi
 # verify device is not currently using aufs. Otherwise must migrate first.
 if $(version_gt "${target_version}" "${aufs_removed_version}") ||
       [ "$target_version" = "$aufs_removed_version" ] &&
-      $(balena info |grep "Storage Driver: aufs"); then
+      balena info 2>/dev/null |grep -q "Storage Driver: aufs"; then
     log ERROR "This device is using the AUFS storage driver and must be migrated before upgrading to v${aufs_removed_version} or later. \
 Upgrade to a release between v${aufs_minimum_migration} and < v${aufs_removed_version}."
 fi
--- a/upgrade-2.x.sh
+++ b/upgrade-2.x.sh
@@ -14,6 +14,7 @@ set -o pipefail
 minimum_hostos_version=2.14.0
 minimum_target_version=2.16.0
 minimum_supervisor_stop=2.53.10
+minimum_aufs_unsupported=8.0.0
 
 # This will set VERSION, SLUG
 # shellcheck disable=SC1091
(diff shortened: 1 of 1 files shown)


upgrade-2.x.sh
```diff
@@ -14,6 +14,7 @@ set -o pipefail
 minimum_hostos_version=2.14.0
 minimum_target_version=2.16.0
 minimum_supervisor_stop=2.53.10
+minimum_aufs_unsupported=8.0.0
```
klutchell:
```suggestion
minimum_aufs_migration=2.85.0
maximum_aufs_migration=8.0.0
```

upgrade-2.x.sh
```diff
@@ -1078,6 +1079,12 @@ if [ "${SLUG}" = "raspberrypi4-64" ] && \
     fi
 fi
 
+if version_gt "${target_version}" "${minimum_aufs_unsupported}" || \
+    [ "$target_version" == "$minimum_aufs_unsupported" ] && \
+    $(balena info |grep "Storage Driver: aufs"); then
+    log ERROR "BalenaEngine aufs storage supported only before ${minimum_aufs_unsupported}; first upgrade to at least 2.83 to convert"
```
klutchell:
```suggestion
    log ERROR "This device is using the AUFS storage driver and must be migrated before upgrading to v${maximum_aufs_migration} or later; Upgrade to a release between v${minimum_aufs_migration} and < v${maximum_aufs_migration}"
```

upgrade-2.x.sh
```diff
@@ -1078,6 +1079,12 @@ if [ "${SLUG}" = "raspberrypi4-64" ] && \
     fi
 fi
 
+if version_gt "${target_version}" "${minimum_aufs_unsupported}" || \
+    [ "$target_version" == "$minimum_aufs_unsupported" ] && \
```
klutchell:
```suggestion
    [ "$target_version" == "$maximum_aufs_migration" ] && \
```

upgrade-2.x.sh
```diff
@@ -1078,6 +1079,12 @@ if [ "${SLUG}" = "raspberrypi4-64" ] && \
     fi
 fi
 
+if version_gt "${target_version}" "${minimum_aufs_unsupported}" || \
```
klutchell:
```suggestion
if version_gt "${target_version}" "${maximum_aufs_migration}" || \
```

upgrade-2.x.sh
kb2ma: I agree "aufs_unsupported" is hard to grok. I agree best not to hard code the migration version later in the script. OTOH `maximum_aufs_migration` is not accurate. It is difficult to specify the last version containing the feature. I rename the variable in the next commit.

upgrade-2.x.sh
kb2ma: I add the longer message in the next commit.

upgrade-2.x.sh
kb2ma: See discussion above on variable names.

upgrade-2.x.sh
kb2ma: Resolved in next commit.

upgrade-2.x.sh
```diff
@@ -14,6 +14,12 @@ set -o pipefail
 minimum_hostos_version=2.14.0
 minimum_target_version=2.16.0
 minimum_supervisor_stop=2.53.10
+# Metabalena OS version at which aufs support was removed from the included balenaEngine,
+# and minimum version that provides migration to overlayfs.
+# Provides an interval of versions to which such a device must hop before updating
+# to latest release.
+aufs_removed_version=8.0.0
```
kb2ma: Variable name is accurate and should be clearer than "minimum_aufs_unsupported".

upgrade-2.x.sh
```diff
@@ -1044,28 +1050,40 @@ fi
 
 if [ -n "$target_version" ]; then
     case $target_version in
-        [2-9].*|2[0-9][0-9][0-9].*.*)
-        if ! version_gt "$target_version" "$minimum_target_version" &&
-            ! [ "$target_version" == "$minimum_target_version" ]; then
+        2[0-9][0-9][0-9].*.*)
```
kb2ma, line 1052: Refactored this section to break out ESR handling. Simplest to overwrite variables here rather than introduce another pair of variables.

upgrade-2.x.sh
```diff
@@ -1044,28 +1050,40 @@ fi
...
+            # Metabalena ESR OS version at which aufs support was removed from
+            # the included balenaEngine, and minimum version that provides migration
+            # to overlayfs.
+            # Overwrites the variable values defined at the top of the script,
+            # which are based on rolling OS version.
+            aufs_removed_version=2026.7.0
+            aufs_minimum_migration=2022.1
+            ;;
+        [2-9].*)
+            # Note: Update comparison to include ESR target when minimum target
```
kb2ma, line 1060: A little PR topic drift here, but accurate. I know we plan to further increase the minimum target soon, but this section likely will see more work anyway as we improve ESR support. This check used to include ESR target, but it was useless because minimum target only is expressed as rolling.

upgrade-2.x.sh
```diff
@@ -1044,28 +1050,40 @@ fi
...
-                    log ERROR "Target OS version \"$target_version\" must be greater than current version."
-                fi
-                log "Target OS version \"$target_version\" OK."
             fi
             ;;
         *)
             log ERROR "Target OS version \"$target_version\" not supported."
             ;;
     esac
+
```
kb2ma, line 1071: Had to move below case so handled by both rolling/metabalena and ESR versioning.

kb2ma:
@klutchell , would appreciate another look. Haven't tested yet so still in draft, but I'm happy with the update.

* Fixup variable names
* Add ESR support
* Refactor target version validation

upgrade-2.x.sh
```diff
@@ -1044,28 +1050,40 @@ fi
 
 if [ -n "$target_version" ]; then
     case $target_version in
-        [2-9].*|2[0-9][0-9][0-9].*.*)
-        if ! version_gt "$target_version" "$minimum_target_version" &&
-            ! [ "$target_version" == "$minimum_target_version" ]; then
+        2[0-9][0-9][0-9].*.*)
+            # Metabalena ESR OS version at which aufs support was removed from
```
klutchell:
```suggestion
            # meta-balena ESR OS version at which aufs support was removed from
```
NIT but may make grep difficult otherwise

upgrade-2.x.sh
```diff
@@ -14,6 +14,12 @@ set -o pipefail
 minimum_hostos_version=2.14.0
 minimum_target_version=2.16.0
 minimum_supervisor_stop=2.53.10
+# Metabalena OS version at which aufs support was removed from the included balenaEngine,
```
klutchell:
```suggestion
# meta-balena OS version at which aufs support was removed from the included balenaEngine,
```
NIT but may make grep difficult otherwise

upgrade-2.x.sh
```diff
@@ -1044,28 +1050,40 @@ fi
...
+        2[0-9][0-9][0-9].*.*)
+            # Metabalena ESR OS version at which aufs support was removed from
+            # the included balenaEngine, and minimum version that provides migration
+            # to overlayfs.
+            # Overwrites the variable values defined at the top of the script,
+            # which are based on rolling OS version.
+            aufs_removed_version=2026.7.0
+            aufs_minimum_migration=2022.1
+            ;;
+        [2-9].*)
```
klutchell, line 1056: Can we future proof this a bit to allow 2-99?

balenaOS v10 isn't that far down the road and this match only allows single digits.

kb2ma: I agree but would like to defer and bundle with some other generic improvements ([project](https://balena.fibery.io/Work/Project/HUP-script-maintenance-2505)).

upgrade-2.x.sh
kb2ma: Updated in 25a4dde

upgrade-2.x.sh
kb2ma: Updated in 25a4dde

upgrade-2.x.sh
kb2ma: Updated in 25a4dde

upgrade-2.x.sh
kb2ma: Updated in 25a4dde

kb2ma:
@klutchell , I think this is as far as we can take this work until release of engine v25 is imminent since we need the actual rolling OS version and sanity check the ESR version. If it looks OK, I will squash but leave in draft.

In this update:
* Fix actual `balena info | grep` command
* Move rolling variable definition near ESR definitions; not needed at top of script

Tested extensively by lowering 'aufs removed' version. Tested over actual aufs -> overlay2 migrations.

Push `balena info` stderr stream to `/dev/null` to avoid warnings like below in HUP log:
```
WARNING: API is accessible on http://0.0.0.0:2375 without encryption.
         Access to the remote API is equivalent to root access on the host. Refer
         to the 'Docker daemon attack surface' section in the documentation for
         more information: https://docs.docker.com/go/attack-surface/
```

kb2ma:
One last question -- do you consider this update a minor update rather than a patch? It seems more of an enabler for engine v25 than a bug fix.

klutchell:
> One last question -- do you consider this update a minor update rather than a patch? It seems more of an enabler for engine v25 than a bug fix.

I think keep it as a patch for now, it's not really adding anything to the interface. Just a restriction on an OS version that doesn't exist yet.

upgrade-2.x.sh
```diff
@@ -1095,10 +1095,19 @@ fi
 # If target OS version uses a balenaEngine that does not support aufs storage,
 # verify device is not currently using aufs. Otherwise must migrate first.
 if $(version_gt "${target_version}" "${aufs_removed_version}") ||
-      [ "$target_version" = "$aufs_removed_version" ] &&
-      balena info 2>/dev/null |grep -q "Storage Driver: aufs"; then
-    log ERROR "This device is using the AUFS storage driver and must be migrated before upgrading to v${aufs_removed_version} or later. \
+      [ "$target_version" = "$aufs_removed_version" ]; then
+    # Verify 'balena info' succeeds before piping to grep. Collects error output after
```
kb2ma: Below is the warning output from `balena info` I'd like to avoid printing systematically.

```
WARNING: API is accessible on http://0.0.0.0:2375 without encryption.
         Access to the remote API is equivalent to root access on the host. Refer
         to the 'Docker daemon attack surface' section in the documentation for
         more information: https://docs.docker.com/go/attack-surface/
```

kb2ma: @klutchell Added test of `balena info` result separately as a defensive measure. The previous commit piped the info output directly to grep. In one test I saw the script pass this test when it should have found aufs storage.

LMK if you have any concerns with these changes.

upgrade-2.x.sh
```diff
@@ -1078,6 +1092,24 @@ if [ "${SLUG}" = "raspberrypi4-64" ] && \
     fi
 fi
 
+# If target OS version uses a balenaEngine that does not support aufs storage,
+# verify device is not currently using aufs. Otherwise must migrate first.
+if $(version_gt "${target_version}" "${aufs_removed_version}") ||
```
klutchell:
```suggestion
if version_gt "${target_version}" "${aufs_removed_version}" ||
```

upgrade-2.x.sh
```diff
@@ -1078,6 +1092,24 @@ if [ "${SLUG}" = "raspberrypi4-64" ] && \
...
+      [ "$target_version" = "$aufs_removed_version" ]; then
+    # Verify 'balena info' succeeds before piping to grep. Collects error output after
+    # command fails to avoid side effect of logging confusing warnings on success as well."
+    if info_out=$(balena info 2>/dev/null); then
+        if $(echo "${info_out}" |grep -q "Storage Driver: aufs"); then
```
klutchell:
```suggestion
        if grep -q "Storage Driver: aufs" <<<"${info_out}"; then
```
Remove parenthesis for exit code only commands.

upgrade-2.x.sh
```diff
@@ -1078,6 +1092,24 @@ if [ "${SLUG}" = "raspberrypi4-64" ] && \
...
+            log ERROR "This device is using the AUFS storage driver and must be migrated before upgrading to v${aufs_removed_version} or later. \
+Upgrade to a release between v${aufs_minimum_migration} and < v${aufs_removed_version}."
```
klutchell:
```suggestion
Upgrade to a release between ${aufs_minimum_migration} and < ${aufs_removed_version}."
```
I just noticed that to remain consistent with the rest of the script, we should drop the v prefixes.

upgrade-2.x.sh
```diff
@@ -1078,6 +1092,24 @@ if [ "${SLUG}" = "raspberrypi4-64" ] && \
...
 
+# If target OS version uses a balenaEngine that does not support aufs storage,
+# verify device is not currently using aufs. Otherwise must migrate first.
+if $(version_gt "${target_version}" "${aufs_removed_version}") ||
+      [ "$target_version" = "$aufs_removed_version" ]; then
+    # Verify 'balena info' succeeds before piping to grep. Collects error output after
+    # command fails to avoid side effect of logging confusing warnings on success as well."
+    if info_out=$(balena info 2>/dev/null); then
+        if $(echo "${info_out}" |grep -q "Storage Driver: aufs"); then
+            log ERROR "This device is using the AUFS storage driver and must be migrated before upgrading to v${aufs_removed_version} or later. \
```
klutchell:
```suggestion
            log ERROR "This device is using the AUFS storage driver and must be migrated before upgrading to ${aufs_removed_version} or later. \
```

upgrade-2.x.sh
```diff
@@ -1078,6 +1092,24 @@ if [ "${SLUG}" = "raspberrypi4-64" ] && \
     fi
 fi
 
+# If target OS version uses a balenaEngine that does not support aufs storage,
+# verify device is not currently using aufs. Otherwise must migrate first.
+if $(version_gt "${target_version}" "${aufs_removed_version}") ||
+      [ "$target_version" = "$aufs_removed_version" ]; then
+    # Verify 'balena info' succeeds before piping to grep. Collects error output after
+    # command fails to avoid side effect of logging confusing warnings on success as well."
```
klutchell:
```suggestion
    # command fails to avoid side effect of logging confusing warnings on success as well.
```

upgrade-2.x.sh
klutchell: Ideally we would only run the `balena info` command once, capture both stdout and stderr, then grep to make sure it returned both `Storage Driver:` and NOT `aufs`.

But yes, agree we should not dump needless warnings to the logs unless we get into the failure condition.

upgrade-2.x.sh
kb2ma: There are some examples of a "v" prefix in Supervisor handling, but I agree it's more consistent to remove.

upgrade-2.x.sh
kb2ma: Changed to add "version", as in "upgrading to version ${aufs_removed_version}". It reads better.

upgrade-2.x.sh
kb2ma: Commit de37cb7 collects stderr to a file and echos contents to the log on error.

kb2ma: @klutchell Should be ready for another look. The only pending issue should be the OS version for engine v25.

kb2ma: I also tested various combinations and think the changes are in good shape functionally.

upgrade-2.x.sh
```diff
@@ -1078,6 +1092,26 @@ if [ "${SLUG}" = "raspberrypi4-64" ] && \
...
+# If target OS version uses a balenaEngine that does not support aufs storage,
+# verify device is not currently using aufs. Otherwise must migrate first.
+if version_gt "${target_version}" "${aufs_removed_version}" ||
+      [ "$target_version" = "$aufs_removed_version" ]; then
+    # Verify 'balena info' succeeds before piping to grep. Collect error output
+    # separately to avoid side effect of logging confusing warnings on success as well.
+    info_errfile=$(mktemp)
+    if info_out=$(balena info 2>"${info_errfile}"); then
+        rm -f "${info_errfile}"
+        if echo "${info_out}" |grep -q "Storage Driver: aufs"; then
```
klutchell:
```suggestion
        driver=$(grep "Storage Driver:" <<<"${info_out}" || true)
        if [ -z "${driver}" ]; then
            log ERROR "Could not determine storage driver from 'balena info'."
        elif [[ "${driver}" == *aufs* ]]; then
            log ERROR "This device is using the AUFS storage driver..."
        fi
```

upgrade-2.x.sh
```diff
@@ -1078,6 +1092,26 @@ if [ "${SLUG}" = "raspberrypi4-64" ] && \
     fi
 fi
 
+# If target OS version uses a balenaEngine that does not support aufs storage,
+# verify device is not currently using aufs. Otherwise must migrate first.
+if version_gt "${target_version}" "${aufs_removed_version}" ||
```
klutchell, line 1101:
```suggestion
if [ -n "${aufs_removed_version}" ] && version_gt "${target_version}" "${aufs_removed_version}" ||
```
NIT, just future-proofing in case someone breaks the case-statement above. We should also init `aufs_removed_version=""` before the case statement to be safe.

upgrade-2.x.sh
kb2ma: Elegant way to ensure storage driver is defined as well as check for the AUFS driver.

upgrade-2.x.sh
kb2ma: Added a [ -z ] check above the 'if' instead. The [ -n ] check suggestion would have silently skipped the test. I don't think that was your intention.

kb2ma:
Resolved all comments although check above on use of `-z` vs. `-n`. Squashed and moved out of draft. Hopefully we're there.

klutchell (approved):

kb2ma:
Enjoyed the process @klutchell and learned a few more bashisms, too.

PR #429, opened 2026-05-16
upgrade-2.x: do not rely on hostapps being the first in the service list

alexgg:
When hostapp extensions are part of the OS composition, the order is not guaranteed to contain the hostapp first.

Filter the hostapp to match by service name and fallback to the first/only entry.

Change-type: patch


upgrade-2.x.sh
```diff
@@ -640,19 +640,31 @@ function hostapp_based_update {
...
+            // (if ($imgs | length) == 1 then $imgs[0].image[0] else empty end) )
+        | select(. != null)
+        | "\(.is_stored_at__image_location)@\(.content_hash)"]'
+
     image=$(CURL_CA_BUNDLE="${TMPCRT}" ${CURL} \
         -H "Content-Type: application/json" \
         -H "Authorization: Bearer ${APIKEY}" \
-        "${API_ENDPOINT}/v6/release?\$select=id&\$expand=contains__image/image&\$filter=(belongs_to__application/any(a:a/is_for__device_type/any(dt:dt/slug%20eq%20'${SLUG}')%20and%20is_host%20eq%20true))%20and%20is_invalidated%20eq%20false%20and%20raw_version%20eq%20'${version}'" \
-        | jq -r "[.d[] | .contains__image[0].image[0] | [.is_stored_at__image_location, .content_hash] | \"\(.[0])@\(.[1])\"]")
+        "${API_ENDPOINT}/v6/release?${expand_query}&\$filter=(belongs_to__application/any(a:a/is_for__device_type/any(dt:dt/slug%20eq%20'${SLUG}')%20and%20is_host%20eq%20true))%20and%20is_invalidated%20eq%20false%20and%20raw_version%20eq%20'${version}'" \
```
klutchell: We should switch to v7 while we are here

upgrade-2.x.sh
```diff
@@ -640,19 +640,31 @@ function hostapp_based_update {
...
     # we need to strip the target_version's variant tag to query the API properly
     local version=${1/.dev/}
     version=${version/.prod/}
 
     # TODO: Get the target variant from the raw version the user provided
     variant_tag=$(echo "${VARIANT:-production}" | tr "[:upper:]" "[:lower:]")
 
+    # Host OS releases can bundle several images so the hostapp is
+    # not guaranteed to be first in the list.
+    expand_query='$select=id&$expand=contains__image($expand=image($select=is_stored_at__image_location,content_hash;$expand=is_a_build_of__service($select=service_name)))'
```
kb2ma, line 641: Any particular reason to add the $select on the objects? It makes sense they reduce data transfer, but just wondering if another reason.

kb2ma:
The overall approach looks OK to me @alexgg . It seems nicely factored. I'm walking through the changes and testing.

I tagged @thgreasi for a review for Pine expertise.

kb2ma:
@alexgg I have some concerns with the hostapp_selector. I generally find `jq` difficult to follow, and this construction is in that category. How did you generate it?
```
hostapp_selector='[.d[]
        | (.contains__image // []) as $imgs
        | ( ($imgs | map(.image[0]) | map(select(.is_a_build_of__service[0].service_name == "hostapp"))[0])
            // (if ($imgs | length) == 1 then $imgs[0].image[0] else empty end) )
        | select(. != null)
        | "\(.is_stored_at__image_location)@\(.content_hash)"]'
```

I tested with the file attached below, which is missing a 'hostapp' image, and it responded with the image from the 'kernel-modules-raspberrypi4-64' image.

Then I tried my hand at the selector and came up with below. I think this is easier to read and returns an empty array with that incomplete JSON output file.
```
'[.d[] | .contains__image | map(.image[0]) | map(select(.is_a_build_of__service[0].service_name == "hostapp"))[0] as $img
       | if $img != null then $img else empty end |"\(.is_stored_at__image_location)@\(.content_hash)" ]'
```

[response-missing-hostapp.txt](https://github.com/user-attachments/files/30289153/response-missing-hostapp.txt)

kb2ma:
I also used the Pine query below to generate release test data.

[pine-query-release-pr429.txt](https://github.com/user-attachments/files/30289391/pine-query-release-pr429.txt)

upgrade-2.x.sh
alexgg: hey @kb2ma, just data transfer. The $expand is what the selector needs, and the $select on the expanded image keeps the response to the two fields used.

alexgg:
> @alexgg I have some concerns with the hostapp_selector. 

hey @kb2ma indeed, I used real OS release data from a subset of releases, but they all contains either a `main` or a `hostapp`, so I did not consider the hypothetical case where there would be no hostapp.

Your proposal does not work for releases with the legacy `main`, or when they miss a container image. I have adapted it to match the literals `hostapp` and `main`.

> I generally find `jq` difficult to follow, and this construction is in that category. How did you generate it?

Indeed, it was trial an error and extensive testing and tweaking. I wasn't going to, but in the end I have converted the data I used to manually test and added it as bats tests that run on CI. That should make it easier to review functionally.

alexgg:
Pending manual revalidation with the latest changes.

kb2ma:
The reworked hostapp selector is easier to understand and passes the tests I had been using. I was not aware of the use of "main" for the hostapp service.

Thanks for taking the time with bats! Unit testing has been sorely needed. The test suite ran for me. I will take a closer look to be sure no issues with moving the variable definitions in the upgrade script itself.

upgrade-2.x.sh
```diff
@@ -806,11 +806,33 @@ function post_update_fixes() {
...
+
+# This will set VERSION, SLUG
+# shellcheck disable=SC1091
+. /etc/os-release
+
+# Don't run anything before this source as it sets PATH here
+# shellcheck disable=SC1091
+source /etc/profile
+
+# workaround for self-signed certs, waiting for https://github.com/balena-os/meta-balena/issues/1398
```
kb2ma, line 829: I reviewed this issue and balena-os/meta-balena#1969, which indicates it should be possible to remove this code. I don't want to slow down this PR, so added review to a script maintenance project I expect to work on soon.

kb2ma:
Continuing to make progress. I reviewed the bats framework and implementation here. No concerns.

Next step is to review some specific queries especially with tags to confirm accuracy.

tests/stubs/releases/curl
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
kb2ma, line 1: This file probably should be in a `stubs/releases` sub-directory since it is specific to this data. It does have quite a bit of common code though. Maybe can be re-factored in the future when the need arises.

kb2ma:
I just pushed up a [kb2ma/oci-runtime](https://github.com/balena-os/balenahup/tree/kb2ma/oci-runtime) branch. It adds a realistic test for the second query -- when the first query result is empty. I think this is the main reason for the second query -- old ESR releases like RPi3 v2021.10.2. Notice it does not work with the first query because _raw_version_ is `0.0.0+rev15`.

If you agree, please add to this PR.

Here is the pine version of the second query: [pine-query-fallback-release-pr429.txt](https://github.com/user-attachments/files/30492458/pine-query-fallback-release-pr429.txt)

kb2ma:
Once we resolve these last two comments, I am ready to approve. I am going to sleep so much better knowing these tests are here. This work is awesome, @alexgg.

tests/stubs/releases/curl
alexgg: Done. We'll defer the refactor.

alexgg:
> If you agree, please add to this PR.

Done.

klutchell (approved):

kb2ma:
Look great to me. I can't approve since I have a commit on the PR now.

PR #438, opened 2026-09-08
Use tmp directory when login to registry

kb2ma:
Workaround for error in OS v6.5-6.6 where HOME directory not valid.

Use of a tmp directory works around a bug in balenaOS v6.5.31 -> 6.6.2 inclusive,
where the HOME directory may be defined incorrectly, and so the login command
cannot cache the password in $HOME/.balena-engine. This problem is significant for
private device types, and where the update is initiated from a transient systemd service by
a v19+ Supervisor.

Change-type: patch


kb2ma:
Tested on an aarch64 private device, raspberrypi3, raspberrypi4-64, and generic-amd64 qemu.

pipex (approved):

